PR #2, opened 2015-04-21
Added Descriptions for All Cards from Dominion Adventures

nickv2002:
Added Descriptions for All Cards from Dominion including Event and
Upgrade cards.


sumpfork:
Awesome, thanks for doing that. Could you change the tags to only have ':::' instead of '::::' (three vs. four colons) as that's the format the script scans for? Thanks.

Also note that it'll still need the changes to the dominion_cards.txt file to actually appear in the output.

nickv2002:
Okay, I removed the `:` (for some reason I thought it was 4 colons).

I'm hoping to find somewhere that has already typed all the card text so that `dominion_cards.txt` doesn't require typing all the text on the cards (which I can't quite read in the [rules PDF](http://riograndegames.com/getFile.php?id=1907) anyway due to the compressed art).

sumpfork:
I think I used to get them from http://dominion.diehrstraits.com/?set=All&f=list (thus the weird format of that file), but that hasn't been updated yet form the looks of it.

nickv2002:
Also not sure if I got all the formatting right.  Not sure how permissive/particular your script is but I can tweak that easily.

nickv2002:
Someone on BGG pointed me to this full text spoilers list, so I'll try to adapt that for the `dominion_cards.txt`
http://www.reddit.com/r/dominion/comments/330n8e/full_adventures_rulebook_spoilers/

Issue #13, opened 2015-05-19
Page and Peasant Cost


sumpfork:
From BGG thread comment: Found an error just now with the Adventures tabs - the upgrades to Page and Peasant should have an asterisk in their cost.

selfthinker:
It seems only Prizes have an asterisk in their cost. It is also missing on Peddler and Spoils.
All the pluses are missing from various Guilds cards as well.

sumpfork:
Agreed, asterisk option was not available before and the card db needs a pass to add them.

sumpfork:
Thanks! Merged and deployed to online generator.

PR #19, opened 2015-07-27
Issue 13 - Updated costs with * and +

achapp:
Issue #13: Page and Peasant Cost

Fixed costs to add * and + where necessary. Modified regex to allow 
cost to end in * or +. Adventures cards missing from yaml version so
you'll need to keep this in mind when those are added.

Tested against both yaml and non-yaml card databases.


sumpfork:
You should be able to update the yaml by exporting as yaml.

PR #27, opened 2015-10-16
Enhancements - fulltab, center_name, back_offset_height

wvoigt:
Added 3 enhancements:
--fulltab to make the tab the full width of the divider
--center_name to center the card name on the tab
--back_offset_height to add points in the vertical direction on the back
page

### DIFF
--- a/dominion_tabs.py
+++ b/dominion_tabs.py
@@ -450,8 +450,11 @@ class DominionTabs:
                     h -= h / 2
 
             words = line.split()
-            if rightSide or not self.options.edge_align_name:
-                w = textInset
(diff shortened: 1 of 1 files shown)


sumpfork:
Thanks, that's great.

Though I do feel that we should probably replace the various name alignment options with a single one that has several choices, "left", "edge" and "centre".

wvoigt:
and center :-)  Any worry about backwards compatability with the options?

sumpfork:
It looks like I haven't exposed any of them on the generator page at http://domtabs.sandflea.org, and combining them would actually make that easier. So no, no backwards compatibility problems that I can think of right now.

wvoigt:
OK, I'll work on that.

PR #28, opened 2015-10-17
Consolidation of options

wvoigt:
Consolidated options on tab location and tab text alignment.  Options
that were generally mutually exclusive to each other were combined.

Replaced --center_name and --edge_align_name with single
--tab_name_align that takes the parameters left, right, center, centre,
edge (default=left)

Replaced --samesidelables and --fulltab with --tab_side that takes the
parameters left, right, alternate, full (default=alternate)

### DIFF
--- a/dominion_tabs.py
+++ b/dominion_tabs.py
@@ -554,6 +554,7 @@ class DominionTabs:
         elif self.options.tab_side == "left" or self.options.tab_side == "full":
             rightSide = not useExtra
         else:
+            # alternate the cards
             if not useExtra:
                 rightSide = not self.odd
             else:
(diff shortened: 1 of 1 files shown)


wvoigt:
I just thought of one more option in this vain.  Change alternate (which always starts with a tab on the right) to right-alternate and left-alternate which specifies the side of tab for the first divider.  Default would be right-alternate.

Between these options, someone can pretty much generate just about any variation of tabs possible.
If you are OK with that, I'll start that next.

sumpfork:
Sounds worth a try.

Peter

wvoigt:
OK, I've made the updates.  Hopefully that makes the website easier.  I think with just two pull-downs selections, you can now cover all cases of tab location on the cards and card name location on the tabs.  If you have any questions, please let me know.

wvoigt:
I realize I broke the current website for "Same Side Labels". If you want, I can put the optoin --samesidelabels back in and alias it to --tab_side right.   Then you can make website changes at any time.

Otherwise, you will either need to
(a) implement the two selection pull-downs (and in that case you would no longer need the "Same Side Labels" checkbox.)
or (b) at a minimum, change the actual option name used for the "Same Side Labels" checkbox to be --tab_side right instead of --samesidelabels

sumpfork:
Leave it, I'll make the updates when I pull this change tithe web page.

PR #33, opened 2015-11-24
Added options to select the text for the front and back of the divider.  Also fixes #31

### LINKED ISSUE #31: One sided version with rules
From BGG user InfoCynic:
```
Is it possible to generate:
* 1-sided
* Omit card text
* Include rules text 

That way if we're printing on one-sided labels or taping a lightweight paper to something heavier, we
have the important info (the rules text). We can always read the cards if we want to know what they
say.```

wvoigt:
1. Added --front and --back to select the type of text to display.
Options are 'card' for the card text, 'rules' for the extra rules,
'blank' to not print anything.  And for --back, the option 'none' will
keep the pages with the backs from printing.
2. Added stub test cases for card text as well as tab size and text
alignment.
3. removed options --no-card-rules and --no-card-backs since they are no
longer needed.
4. Fixed issue with the tab outline drawing
5. Fixed issue when tab_name_align was being assigned 'left'


wvoigt:
Wow, even as simple as they are, the tests actually found a coding error.  This is cool.
Is there a way to run these tests locally first?  I could not figure out how to do that.

sumpfork:
Easiest is to install nose and run `nosetests` in the to level dir - that's what happens on CI now.

PR #44, opened 2016-04-25
Pull request for adding French to Dominion Tabs Generator

piroxxi:
Hi,

I've started creating a french translation for Dominion's cards. For now there is only card's names translated. I'll try to see if I can find an online database of the cards descriptions.

Please tell me if something seems wrong,

Regards,
Raphaël


piroxxi:
Hi,

I've succeeded in generating a PDF file with french titles using python directly. I had to correct the cards.json file. There still is some translations missing.

PR #51, opened 2016-06-08
Initial support for Empires

splatteredbits:
I've added some initial support for Empires: I've added the cards with their cost, name, and type. I added a description for some before I ran out of time. No artwork. No support for debt. I only use text on my dividers and only put the name of the card and its expansion. Hope this gets you started.

Also, fixed a bug with the --use-text-set-icon parameter.

### DIFF
--- a/domdiv/cards.py
+++ b/domdiv/cards.py
@@ -90,10 +90,10 @@ class Card(object):
 
     @classmethod
     def getSetText(cls, setName, cardName):
-        if setName in cls.setTextIcons:
-            return cls.setTextIcons[setName]
-        if cardName.lower() in cls.promoTextIcons:
-            return cls.promoTextIcons[cardName.lower()]
+        if setName in setTextIcons:
+            return setTextIcons[setName]
+        if cardName.lower() in promoTextIcons:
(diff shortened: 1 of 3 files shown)


domdiv/cards.py
```diff
@@ -193,11 +195,14 @@ def getTabCostHeightOffset(self):
...
     CardType(('Action', 'Ruins'), 'ruins.png', 0, 1),
     CardType(('Action', 'Shelter'), 'action-shelter.png'),
+    CardType(('Action', 'Attack', 'Duration'), 'duration.png'),
     CardType(('Action', 'Attack', 'Looter'), 'action.png'),
     CardType(('Action', 'Attack', 'Traveller'), 'action.png'),
     CardType(('Action', 'Reserve'), 'reserve.png'),
     CardType(('Action', 'Reserve', 'Victory'), 'reserve-victory.png'),
     CardType(('Action', 'Traveller'), 'action.png'),
+    CardType(('Action', 'Gathering'), 'action.png'),
+    CardType(('Action', 'Treasure'), 'treasre.png'),
```
sumpfork: treasre->treasure

PR #52, opened 2016-06-11
More Empires support

bog:
- adds Empires set icon (http://wiki.dominionstrategy.com/index.php/File:Empires_icon.png)
- adds debt graphic (http://wiki.dominionstrategy.com/images/thumb/8/81/Debt.png/38px-Debt.png)
- supports cards with debt cost only. `Fortune` and `Wedding` would appear weird.
- removes trailing whitespace from cards.json


sumpfork:
Thanks, hoping to do a once-over and quick test on the weekend. Sadly very busy, might merge in blindly if I don't find more time.

PR #62, opened 2016-06-15
Added Event and Landmarks.

robertre2:
Added Event and Landmarks from Empires, and the Summon Event.
Added option to disable Landmarks.
Fixed display of Debt in descriptions.


sumpfork:
I'm going to do an untested merge for this and test later - looks like good stuff to have on the main branch for everyone, and I'm too busy to test right now. Will test before release/pushing to online generator.

Issue #45, opened 2016-05-26
Group Special Cards (Shelters) are incorrect color


Bob-A-Mickelson:
When Shelters are grouped the banner label is white and red. the banner label should be solely red as not all shelters are not Action-Shelters. (White-Red) The color all shelters have in common is red. This is also how Dark Ages lists Shelters in the box insert. 

BTW great job on putting together this program.

wvoigt:
I'm in the process of making changes for #47 and #48 and while I'm there I can work on this.

Issue #48, opened 2016-05-28
Option to print dividers as wrappers or slipcases


wvoigt:
I would like to have the option to print the dividers as wrappers / slipcases.  Similar to [hgzwopjp's Dominion slipcases](https://boardgamegeek.com/filepage/64912/hgzwopjps-dominion-slipcases-horizontal-vertical-c).  He has not updated them for the last expansion and has not replied to messages.

To make the wrappers / slipcases, one must know the thickness of the stack of cards.  So this is dependent on #47.

sumpfork:
Note that I have a complete solution for generating card tuckboxes in python that I'd be willing to open source for this if it helps.

wvoigt:
Thanks.  I'm just about done.  My last task is to go through the other language files to fix them up.  I have two comments/questions:

1. Updating the language files is not easy.  It looks like some are missing cards.  I would think there should be a better way to do this.  There is a lot of redundant structure data.  Maybe a primary file, and then the other languages just provide the text changes for that language.  The nice part of that would be that once a card is added to the base language, the card will show up, just not in the other language.  Possibly prompting others to help fix/add. I would think that would take some form of key for each card.  Can't be the name since some are used across sets, but possibly set-name.  I would think that would be unique.
2. The changes for the card count and wrapper/slipcase are pretty extensive.  Did you want them just submitted, or would you prefer them in a separate branch until more eyes can take a look at it?

wvoigt:
You can see the changes now at:
https://github.com/sumpfork/dominiontabs/compare/master...wvoigt:master

sumpfork:
Looks pretty good on first glance. Want to submit a pull request, or should I test it off of your branch first?

wvoigt:
I made a few minor updates to clean it up a little.  I would feel more comfortable if it was tested a little more.  I've run through a number of tests, but a second set of eyes is always good.   It *should* behave the same for all previous options.

wvoigt:
Also note this addresses issues #47 and #45

sumpfork:
Honestly, just make a pull request, otherwise I'll keep forgetting it's in the pipeline and to test it :).

Issue #47, opened 2016-05-28
Add Card counts to database, have option to print on dividers


wvoigt:
I want the option to print the number of cards on the divider, with the following extras:
1. For groupings, if only the one divider is printed, then that displays the sum of all the cards under that grouping.
2. The same goes for Events and Prizes

wvoigt:
I did want to have a discussion on the best (or should I say acceptable) way to add the card counts.  There are a couple of options:
1. The obvious one is to add a
`  "count": xx, `
in every entry in the cards.json.  This would be a lot of work and it must be done for each and every language as well.
2. Noticing that most card types have a consistent number of cards, this could be a quicker way to add the counts across all languages.  However, there are about 10 cards that are exceptions to this.

I'm proposing a hybrid.  Implement (2).  Then for the 10 or so exceptions, add those to the cards.json.  Many fewer mandatory changes to cards.json.  In effect, if a "count" is in the cards.json it is used, otherwise use the default for the card type specified.

This approach also has the bonus of making a transition to (1) easier if that is ever wanted.

Does this sound reasonable?

sumpfork:
Hybrid sounds good to me.

PR #65, opened 2016-06-19
Add card counts and wrapper/slipcase features

wvoigt:
This adds card counts and the option to print wrappers (aka slipcases).
It fixes issues #45, #47, and #48.


wvoigt:
This has passed my testing.   Note that the number of wrappers is not optimized on a page.  It basically assumes that each wrapper could be as big as the largest wrapper (i.e., Copper).  Optimization can only come with a restructuring of how cards are fit on a page.

sumpfork:
Missing `images/card.png`? I think you had it in your previous request, maybe just forgot this time?

wvoigt:
Yes, forgot.  Added back in.

PR #74, opened 2016-10-01
Add German texts for promo cards.

0-wiz-0:

PR #77, opened 2016-10-04
German promo cards

sumpfork:
This is based on #74 but adds the mapping for the promo card names and fixes some of the logic surrounding missing card set images.


0-wiz-0:
Thanks for the fixes! Please merge so I can more easily try working on Sauna next :)

PR #78, opened 2016-10-05
Add German Sauna/Avanto promo cards.

0-wiz-0:
I'm not sure about mapping.json, please check it. Thank you!


sumpfork:
It works for adding the cards, but note that we don't actually have set images for these cards yet.

Issue #73, opened 2016-09-22
How to Deal with Second Edition Cards


nickv2002:
How do you want to deal with the [second edition](https://boardgamegeek.com/article/23765096#23765096) which are already shipping to stores?

Some Options I thought of:
1. New "Second Edition" sets for Dominion and Intrigue
2. New "Second Edition" Group of the new cards

nickv2002:
Additionally, it sounds like we're getting new rules to be updated and new wording on a few cards (adding "may") that might actually change the rules for them slightly.

sumpfork:
I think '2nd ed.' versions of Dominion and Intrigue will probably work ok. These can also have the wording changes. There'll be a good amount of data duplication, obviously - I guess we could get fancy and have sets inherit from others with overriding, adding and deleting, but I'm not sure that's worth the effort and would make it harder for people to contribute that aren't familiar with how those mechanisms work. 

We already have an example of (sub-)set duplication between the 'Dominion' and 'Base' sets.

I'm not sure we'll be able to track individual slight card text/rule updates in new printings of other sets very well, and I suspect it probably doesn't matter that much - people do have the actual rules and cards in case there are questions.

nickv2002:
@stumpfork What about for the folks that just purchase the upgrade packs with the 7 new cards (14 total between the 2 sets)? Can we also put in an _Upgrade_ set for those?

sumpfork:
To just print the new dividers instead of printing a whole set again, you mean? Hmm, yeah, I guess that'd be nice - in fact, I might need that :).

Well, is it worth having some basic inheritance in that case? Like a `set_definitions.json` file that lists something like
```
[
...
{
    "set": "Dominion 2nd. Edition",
    "inherits_from": "Dominion",
    "cards_to_delete": ["cardname1", "cardname2"...]
},
...
]
```
with the default being that cards from parent sets are included if not deleted, and overridden if specified again in the actual `cards.json` file for the inheriting set?

nickv2002:
I'm fine with inheritance. But there's not a _lot_ of crossover so I'm also fine with the dumb option of just having duplicated data for cards that are in multiple sets.

(I figure many people with existing prints of the dividers will be buying the upgrade packs with the 14 new cards, myself included.)

nickv2002:
[...]

PR #83, opened 2016-10-16
added update cards to the dominion set (english only)

mr-ice:
Hey sumpfork,

Thanks for the dominiontabs code, I love the wrappers and I'm currently printing and making the wrappers as I incorporate Empires and my update pack.  (migrating away from tabs only as my fat fingers have trouble extracting the cards from my vertical organizer slots).

This pull request is for adding the 7 cards added in the update pack to the english card_db.  I don't know other languages to help with those.

According to wiki.dominionstrategy.com these are in the base set, so I included them as 'dominion' cards.

### DIFF
--- a/card_db/en_us/cards.json
+++ b/card_db/en_us/cards.json
@@ -4382,5 +4382,84 @@
   "types":[
    "Action"
   ]
+ },
+ {
+  "cardset": "dominion",
+  "cost": "3",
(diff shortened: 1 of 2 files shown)


nickv2002:
@mr-ice Thanks for your contribution we're discussing how to best incorporate these with the original cards back in this thread: https://github.com/sumpfork/dominiontabs/issues/73

But all the JSON text you typed up and formatted will be useful in any situation.

PR #105, opened 2016-11-12
Card db refactoring

nickv2002:
This is the big restructuring we discussed at length in issue https://github.com/sumpfork/dominiontabs/issues/73


sumpfork:
I started looking over this, btw, but will need more time to test and stuff. Super busy, but hopefully next week? Have you guys tested extensively?

wvoigt:
@sumpfork  I've been testing as I go.  I've also been trying to incorporate the feedback from @soudek01.   For the most part, I've been focusing on the structural parts.  I believe some of the text suggestions have been done, but probably not all.  Not sure if I would wait for all the text changes.  I'll keep updating the branch until you merge.

wvoigt:
While the pull was waiting, I just kept going.  But at this point, I think I'm done.    Anything left will be individual formatting in the various languages.  I've done much of it, but I'm sure I missed things here and there.

More eyes on the output would be the best right now.

wvoigt:
What is the best way to upload some pdf's of the files here for folks to review?
I can't seem to upload pdf files.

domdiv/__init__.py
```diff
@@ -175,8 +205,10 @@ def parse_opts(arglist):
         default=-1,
         help="stop generating after this many pages, -1 for all")
     parser.add_argument("--language",
-                        default='en_us',
-                        help="language of card texts")
+                        default=LANGUAGE_DEFAULT,
```
sumpfork: can we add an actual `choices=LANGUAGE_CHOICES` or similar? Right now the script dies with an exception if you specify a non-existent language, and it'd be nice if it provided a better help message.

domdiv/draw.py
```diff
@@ -25,19 +26,21 @@ def __init__(self):
         self.canvas = None
 
     def registerFonts(self):
-        try:
-            dirn = os.path.join(self.options.data_path, 'fonts')
+        dirn = os.path.join(self.options.data_path, 'fonts')
+        regularMpath = os.path.join(dirn, 'MinionPro-Regular.ttf')
+        boldMpath = os.path.join(dirn, 'MinionPro-Bold.ttf')
+        obliqueMpath = os.path.join(dirn, 'MinionPro-Oblique.ttf')
```
nickv2002: This was a change based on the names downloaded from FontsGeek but it appears they've renamed to Italic I guess(?)  I will change it from `MinionPro-Oblique` back to `MinionPro-It`

sumpfork: Ah ok. Not for this PR, but I think we should integrate a font copy/download and conversion script, if only to guarantee the right names. Pretty sure we're ok copying/downloading them given that they're available. I checked the license, and we're definitely not allowed to distribute them ourselves.

card_db/en_us/translation.txt
```diff
@@ -0,0 +1,71 @@
+# Translation Instructions
```
wvoigt: I went back and forth on this.  The pain of upkeep vs having the information for a translator all in one directory.  Even the more mature translations still have cards that need translation, so that eliminated just having it in the xx directory.

So it can be moved if we think translators will find it.  Or more likely, remembered if the files are to be sent to someone to translate via email.  It is only documentation, so it can easily be moved.

To be honest, the best place would be to have it in the json files themselves, at the top in comments.  It would take preprocessing the files.  But I understand that is frowned upon.

domdiv/__init__.py
wvoigt: I'll fix this.  Duh (hand plant to forehead.) That was the intent of automatically generating LANGUAGE_CHOICES.  Just forgot to add the important part.

domdiv/__init__.py
```diff
@@ -383,73 +458,194 @@ def baseIndex(self, name):
...
+    set_text_filepath = os.path.join(options.data_path,
+                                     "card_db",
+                                     language,
+                                     "sets_" + language + ".json")
+    with codecs.open(set_text_filepath, 'r', 'utf-8') as set_text_file:
+        set_text = json.load(set_text_file)
+    assert set_text, "Could not load set text for %r" % language
+
+    # Now apply to all the sets
+    for set in sets:
```
wvoigt: I'll fix this.

domdiv/__init__.py
```diff
@@ -383,73 +458,194 @@ def baseIndex(self, name):
...
+        if set in set_text:
+            for key in set_text[set]:
+                sets[set][key] = set_text[set][key]
+    return sets
+
+
+def combine_cards(cards, old_card_type='', new_card_tag='', new_cardset_tag=None, new_type=None):
```
wvoigt: I'll fix this.

domdiv/__init__.py
```diff
@@ -383,73 +458,194 @@ def baseIndex(self, name):
...
+    filteredCards = []
+    holder = None
+    for c in cards:
+        if c.isType(old_card_type):
+            if holder is None:
```
wvoigt: I'll fix this.

domdiv/__init__.py
wvoigt: This should now be fixed in commit e947c71

domdiv/__init__.py
wvoigt: Fixed with commit e947c71

domdiv/__init__.py
wvoigt: Fixed with commit e947c71

sumpfork:
I might've missed this in the discussion, but it's certainly not obvious from trying to use the script: How am I supposed to just print dividers for the dominion set now? `--expansions dominion` doesn't work because there's no such set name anymore, and `--expansions dominion1stEdition` doesn't work because the code calls `lower()` on it, but the tag has uppercase letters in the db. Also, how is the new edition flag supposed to work with that if I have to specify the edition in the set name already (or do I not somehow?) I would expect to be able to do `--expansions dominion` and then `--edition 1` and similar. Also, it'd be really nice if we could include the possible expansion tags in the help and as choices for the `--expansions` option. Or maybe I'm missing something? At least needs better documentation if I am...

wvoigt:
That's a bug.  While it seems obvious now, none of my test cases explicitly called out just the dominion or intrigue sets.

The `--edition` option is currently used to narrow down the set lists.  And as you found, that does not help if you can't use `--expansions` to add the sets.

I'm thinking the following should change:

- The set key must be all lower case.  The `lower()` really does make matching things easier if any form of capitalization is used on input.  To make it easier to read (and type on the command line), I would propose changing `dominion1stEdition` to `dominion1` and `dominion2ndEditionUpgrade` to `dominion2upgrade`.  Similar shortening for the other sets as well.
- Adding aliasing of `--expansions`.  For example, make `--expansion dominion` expand out to be the equivalent of `--expansion dominion1` plus `--expansion dominion2` plus  `--expansion dominion2upgrade`.  Similar for intrigue.  Then the `--edition` and `--upgrade_with_expansion` options could be used to further reduce if wanted. (Should be no change there.)  Note, this does not preclude someone from calling out dominion1 or dominion2, but it will make it easier to include all versions.

I don't think the aliasing would take that much code if done done early after the options are obtained and should not be too disruptive.  The only downside is some hard coding of these values.
The changing of the set name will take a little more work, simply because `sets_db.json` and the sets_xx.json for each language must be changed.  But overall very manageable. 

Thoughts on this approach?

domdiv/__init__.py
```diff
@@ -383,73 +487,263 @@ def baseIndex(self, name):
...
+            if 'set_name' in Card.sets[card.cardset_tag].keys():
+                card.cardset = Card.sets[card.cardset_tag]['set_name']
+
+    # If expansion names given, then remove any cards not in those expansions
+    # Expansion names can be the names from the language or the cardset_tag
     if options.expansions:
+        options.expansions = list(set(options.expansions))
         options.expansions = [o.lower() for o in options.expansions]
-        reverseMapping = {v: k for k, v in Card.language_mapping.iteritems()}
+        reverseMapping = {set_tag.lower(): Card.sets[set_tag]['set_name'].lower for set_tag in Card.sets}
```
sumpfork: `lower` instead of `lower()`, leads to: 

```Traceback (most recent call last):
  File "dominion_dividers.py", line 6, in <module>
    domdiv.main(sys.argv[1:], os.path.dirname(__file__))
  File "/Users/pgorniak/Dropbox/boardgames/dominiontabs/domdiv/__init__.py", line 967, in main
    return generate(options, data_path)
  File "/Users/pgorniak/Dropbox/boardgames/dominiontabs/domdiv/__init__.py", line 947, in generate
    cards = filter_sort_cards(cards, options)
  File "/Users/pgorniak/Dropbox/boardgames/dominiontabs/domdiv/__init__.py", line 757, in filter_sort_cards
    if c.cardset.lower().startswith(e)), None):
  File "/Users/pgorniak/Dropbox/boardgames/dominiontabs/domdiv/__init__.py", line 757, in <genexpr>
    if c.cardset.lower().startswith(e)), None):
TypeError: startswith first arg must be str, unicode, or tuple, not builtin_function_or_method
```

wvoigt: I know.  I found it this morning too, but did not have time to fix before work.  I'll do a quick update to fix this.  But after looking at this code again, it still has problems.  For example, it will show an error (but work) if you enter 'dominion' since none of the sets are actually called just 'dominion' any more.  

So I hope to have a real update for this part of the code late Friday or on Saturday morning.

sumpfork: great, thanks

wvoigt: @sumpfork  As an aside, are you ok with updating the behavior of `--expansions` to include multiple sets with the same `--expansions` flag.  For example, `--expansions dominion empires`.  It can still take multiple `--expansions` flags as it does today.   And while at it, how about having the alternative spelling  `--expansion` as well.  I can't tell you how many times I have forgot that extra 's'.

I can add with the update Friday/Saturday as well if you are OK.

sumpfork: Both sound good, it is confusing as it stands.

wvoigt:
@sumpfork I fixed the --expansions options.  It should behave correctly now for partial matches.  And I think the code is more efficient (and easier to read too IMHO.) 

I also spent some time organizing the options into groups that I thought made sense together.  It's not perfect, but hopefully a start.   Certainly willing to make other changes (or even change it back).  Part of the change was changing the argparse formatter to
`formatter_class=argparse.ArgumentDefaultsHelpFormatter`.  It automatically displays the defaults.
Captured output below:
```
usage: dominion_dividers.py [-h] [--outfile OUTFILE] [--papersize PAPERSIZE]
                            [--language {de,en_us,fr,it,nl_du}]
                            [--orientation {horizontal,vertical}]
                            [--size SIZE] [--sleeved]
                            [--order {expansion,global,colour,cost}]
                            [--front {card,rules,blank}]
                            [--back {card,rules,blank,none}] [--count]
                            [--types]
                            [--tab_side {left,right,left-alternate,right-alternate,centre,full}]
                            [--tab_name_align {left,right,centre,edge,center}]
                            [--tabwidth TABWIDTH] [--cost {tab,body-top,hide}]
                            [--set_icon {tab,body-top,hide}]
                            [--no-tab-artwork] [--use-text-set-icon]
                            [--expansion_dividers]
                            [--centre_expansion_dividers]
                            [--expansion_dividers_long_name]
                            [--expansions [EXPANSIONS [EXPANSIONS ...]]]
                            [--edition {1,2,latest,all}]
                            [--upgrade_with_expansion]
                            [--base_cards_with_expansion]
                            [--special_card_groups] [--include_blanks]
                            [--exclude_events] [--exclude_landmarks]
                            [--wrapper] [--thickness THICKNESS]
                            [--sleeved_thick] [--sleeved_thin]
                            [--notch_length NOTCH_LENGTH] [--notch]
                            [--minmargin MINMARGIN] [--cropmarks]
                            [--linewidth LINEWIDTH]
                            [--back_offset BACK_OFFSET]
                            [--back_offset_height BACK_OFFSET_HEIGHT]
                            [--vertical_gap VERTICAL_GAP]
                            [--horizontal_gap HORIZONTAL_GAP]
                            [--no-page-footer] [--num_pages NUM_PAGES]
                            [--tabs-only] [--cardlist CARDLIST] [--write_json]

Generate Dominion Dividers

optional arguments:
  -h, --help            show this help message and exit

Basic Divider Options:
  Basic choices for the dividers.

  --outfile OUTFILE, -o OUTFILE
                        The output file name. (default: dominion_dividers.pdf)
  --papersize PAPERSIZE
                        The size of paper to use; '<%f>x<%f>' (size in cm), or
                        'A4', or 'LETTER'. If not specified, it will default
                        to system defaults, and if the system defaults are not
                        found, then to 'LETTER'. (default: None)
  --language {de,en_us,fr,it,nl_du}, -l {de,en_us,fr,it,nl_du}
                        Language of divider text. (default: en_us)
  --orientation {horizontal,vertical}
                        Either horizontal or vertical divider orientation.
                        (default: horizontal)
  --size SIZE           Dimentions of the cards to use with the dividers
                        '<%f>x<%f>' (size in cm), or 'normal' = '9.1x5.9', or
                        'sleeved' = '9.4x6.15'. (default: normal)
  --sleeved             Same as --size=sleeved. (default: False)
  --order {expansion,global,colour,cost}
                        Sort order for the dividers: 'global' will sort by
                        card name; 'expansion' will sort by expansion, then
                        card name; 'colour' will sort by card type, then card
                        name; 'cost' will sort by expansion, then card cost,
                        then name. (default: expansion)

Divider Body:
  Changes what is displayed on the body of the dividers.

  --front {card,rules,blank}
                        Text to print on the front of the divider; 'card' will
                        print the text from the game card; 'rules' will print
                        additional rules for the game card; 'blank' will not
                        print text on the divider. (default: card)
  --back {card,rules,blank,none}
                        Text to print on the back of the divider; 'card' will
                        print the text from the game card; 'rules' will print
                        additional rules for the game card; 'blank' will not
                        print text on the divider; 'none' will prevent the
                        back pages from printing. (default: rules)
  --count               Display card count on body of the divider. (default:
                        False)
  --types               Display card type on the body of the divider.
                        (default: False)

Divider Tab:
  Changes what is displayed on on the Divider Tab.

  --tab_side {left,right,left-alternate,right-alternate,centre,full}
                        Alignment of tab; 'left'/'right' forces all tabs to
                        left/right side; 'left-alternate' will start on the
                        left and then toggle between left and right for the
                        tabs; 'right-alternate' will start on the right and
                        then toggle between right and left for the tabs;
                        'centre' will force all label tabs to the centre;
                        'full' will force all label tabs to be full width of
                        the divider. (default: right-alternate)
  --tab_name_align {left,right,centre,edge,center}
                        Alignment of text on the tab; The 'edge' option will
                        align the card name to the outside edge of the tab, so
                        that when using tabs on alternating sides, the name is
                        less likely to be hidden by the tab in front (edge
                        will revert to left when tab_side is full since there
                        is no edge in that case). (default: left)
  --tabwidth TABWIDTH   Width in cm of stick-up tab (ignored if --tab_side is
                        'full' or --tabs_only is used). (default: 4.0)
  --cost {tab,body-top,hide}
                        Where to display the card cost; may be set to 'hide'
                        to indicate it should not be displayed, or given
                        multiple times to show it in multiple places.
                        (default: ['tab'])
  --set_icon {tab,body-top,hide}
                        Where to display the set icon; may be set to 'hide' to
                        indicate it should not be displayed, or given multiple
                        times to show it in multiple places. (default:
                        ['tab'])
  --no-tab-artwork      Don't show background artwork on tabs. (default:
                        False)
  --use-text-set-icon   Use text/letters to represent a card's set instead of
                        the set icon. (default: False)

Expansion Dividers:
  Adding separator dividers for each expansion.

  --expansion_dividers  Add dividers describing each expansion set. A list of
                        cards in the expansion will be shown on the front of
                        the divider. (default: False)
  --centre_expansion_dividers
                        Centre the tabs on expansion dividers. (default:
                        False)
  --expansion_dividers_long_name
                        Use the long name with edition information on the
                        expansion divider tab. Without this, the shorter
                        expansion name is used on the expansion divider tab.
                        (default: False)

Divider Selection:
  What expansions are used, and grouping of dividers.

  --expansions [EXPANSIONS [EXPANSIONS ...]], --expansion [EXPANSIONS [EXPANSIONS ...]]
                        Limit dividers to only the specified expansions. If no
                        limits are set, then all expansions are included.
                        Expansion names can also be given in the language
                        specified by the --language parameter. Any expansion
                        with a space in the name must be enclosed in quotes.
                        This may be called multiplle times. Values are not
                        case sensitive and can also match the starting
                        characters of an expansion name. For example,
                        'dominion' will match all expansions that start with
                        that name; Choices available in all languages include:
                        adventures, alchemy, cornucopia, dark ages,
                        dominion1stEdition, dominion2ndEdition,
                        dominion2ndEditionUpgrade, empires, guilds,
                        hinterlands, intrigue1stEdition, intrigue2ndEdition,
                        intrigue2ndEditionUpgrade, promo, prosperity, seaside
                        (default: None)
  --edition {1,2,latest,all}
                        Editions to include: '1' is for all 1st Editions; '2'
                        is for all 2nd Editions; 'latest' is for the latest
                        edition for each expansion; 'all' is for all editions
                        of expansions; This can be combined with other options
                        to refine the expansions to include in the output.
                        (default: all)
  --upgrade_with_expansion
                        Include any new edition upgrade cards with the
                        expansion being upgraded. (default: False)
  --base_cards_with_expansion
                        Print the base cards as part of the expansion (i.e., a
                        divider for 'Silver' will be printed as both a
                        'Dominion' card and as an 'Intrigue 1st Edition'
                        card). If this option is not given, all base cards are
                        placed in their own 'Base' expansion. (default: False)
  --special_card_groups
                        Group cards that generally are used together (e.g.,
                        Shelters, Tournament and Prizes, Urchin/Mercenary,
                        etc.). (default: False)
  --include_blanks      Include a few dividers with extra text. (default:
                        False)
  --exclude_events      Group all 'Event' cards across all expansions into one
                        divider. (default: False)
  --exclude_landmarks   Group all 'Landmark' cards across all expansions into
                        one divider. (default: False)

Card Sleeves/Wrappers:
  Generating dividers that are card sleeves/wrappers.

  --wrapper             Draw sleeves (aka wrapper) for the cards instead of a
                        divider for the cards. (default: False)
  --thickness THICKNESS
                        Thickness of a stack of 60 cards (Copper) in
                        centimeters. Typically unsleeved cards are 2.0, thin
                        sleeved cards are 2.4, and thick sleeved cards are
                        3.2. This is only valid with the --wrapper option.
                        (default: 2.0)
  --sleeved_thick       Same as --size=sleeved --thickness 3.2. (default:
                        False)
  --sleeved_thin        Same as --size=sleeved --thickness 2.4. (default:
                        False)
  --notch_length NOTCH_LENGTH
                        Length of thumb notch on wrapper in centimeters (a
                        value of 0.0 means no notch on wrapper). This can make
                        it easier to remove the actual cards from the wrapper.
                        This is only valid with the --wrapper option.
                        (default: 0.0)
  --notch               Same as --notch_length thickness 1.5. (default: False)

Printing:
  Changes how the Dividers are printed.

  --minmargin MINMARGIN
                        Page margin in cm in the form '<%f>x<%f>', left/right
                        x top/bottom). (default: 1x1)
  --cropmarks           Print crop marks on both sides, rather than tab
                        outlines on the front side. (default: False)
  --linewidth LINEWIDTH
                        Width of lines for card outlines and crop marks.
                        (default: 0.1)
  --back_offset BACK_OFFSET
                        Back page horizontal offset points to shift to the
                        right. Only needed for some printers. (default: 0)
  --back_offset_height BACK_OFFSET_HEIGHT
                        Back page vertical offset points to shift upward. Only
                        needed for some printers. (default: 0)
  --vertical_gap VERTICAL_GAP
                        Vertical gap between dividers in centimeters.
                        (default: 0.0)
  --horizontal_gap HORIZONTAL_GAP
                        Horizontal gap between dividers in centimeters.
                        (default: 0.0)
  --no-page-footer      Do not print the expansion name at the bottom of the
                        page. (default: False)
  --num_pages NUM_PAGES
                        Stop generating dividers after this many pages, -1 for
                        all. (default: -1)
  --tabs-only           Draw only the divider tabs and no divider outlines.
                        Used to print the divider tabs on labels. (default:
                        False)

Miscellaneous:
  These options are generally not used.

  --cardlist CARDLIST   Path to file that enumerates each card to be printed
                        on its own line. (default: None)
  --write_json          Write json version of card definitions and extras.
                        (default: False)

Source can be found at 'https://github.com/sumpfork/dominiontabs'. An online
version can be found at 'http://domtabs.sandflea.org/'.
```

card_db/en_us/translation.txt
wvoigt: @sumpfork So I propose a compromise.  I moved the translation instructions to /card_db and removed it from all the *active* languages.  However, I still kept it in the non-active 'xx' language which is the starting point for any new translation.   That way, to start any new language, we only need to give someone (or they grab) every file in that directory and they are ready to start. 

So now the file is in only two places.  Still a little painful to maintain, but hopefully must less of one.  And I'm hoping the file does not change that much as we go forward.  I also update the maintenance tool to copy the latest version into 'xx'.

It might even be good to have a zipped version of the xx directory downloadable somewhere on the webpage.  So someone scarred of GitHub could just download and start.  Hopefully encouraging more languages.

sumpfork:
I don't want to hold this up any longer. I've done a cursory review of the code, and tried a couple of th new options to convince myself that I can integrate this into the online generator. Any further bug fixing can happen on the master branch, as I think this is in general a great idea and don't want to continue accumulating bug reports against old versions.

wvoigt:
Any reason not to delete the branch now?

Issue #131, opened 2016-12-15
Turn this project into a better python package


sumpfork:
Especially for #102, but also for inclusion in other projects and general python standard performance I would like to convert this project into a proper python package.

This means:
1) The top level dir including `setup.py` should not be a package
2) All resources needed (fonts, images, card_db) should be included in the package (this used to work partially, but is broken right now)
3) all resources needed need to be accessible by the the code and the libraries it uses even if they're included in a binary compressed package format
4) There should be no python script to run the generator at the top level of the project. Instead, the script entry point in the setup.py file should be fixed and it should be the only way to run the generator.
5) the version number should move into the `domdiv` package, and that should be the only package installed.

The outcome should be that
1) Running `python setup.py install` makes `dominion_dividers` available as a normally installed python script that works without any further configuration (barring the fact that it'll use non-proprietary fonts by default).
2) Installing the package as in 1) should let another library call the code without any additional configuration.

PR #132, opened 2016-12-15
Make better package

sumpfork:
For #131, and mostly for preparing a better release system for the online generator.

Could you guys have a look to see if it makes sense?

GitHub is terrible at aligning diffs for moved content - main.py is mainly a copy of the old __init__.py, just the parts to do with file reading have changes (look for references to `pkg_resources`).


domdiv/fonts/README.md
```diff
@@ -0,0 +1,4 @@
+#### Fonts
+I believe I cannot distribute one font (Minion Pro) domdiv uses as they are commercially sold by Adobe. However, the font files are included with every install of the free Adobe Reader. For example, on Windows 7 they are in C:\Program Files (x86)\Adobe\Reader 9.0\Resource\Font called `MinionPro-Regular.otf`, `MinionPro-Bold.otf` and `MinionPro-It.otf`.
+
+Sadly, these fonts use features of the otf format that are not support by the reportlab package. Thus, they need to first be converted to ttf (TrueType) format. I used the open source package fontforge to do the conversion. Included as 'convert.ff' is a script for fontforge to do the conversion, on Mac OS X with fontforge installed through macports you can just run `./convert.ff MinionPro-Regular.otf`, `./convert.ff MinionPro-Bold.otf` and `./convert.ff MinionPro-It.otf`. With other fontforge installations, you'll need to change the first line of convert.ff to point to your fontforge executable. I have not done this step under Windows - I imagine it may be possible with a cygwin install of fontforge or some such method. Copy these converted files to this subdirectory.
```
nickv2002: If can we link to a zip of the fonts already bundled up instead?

nickv2002: Even though they're TTF, the FontsGeek MinionPro fonts still have to run through FontForge conversion.  Something about having the right shapes IIRC.

I was thinking finding the converted font files somewhere in the shared hosting for 'off the back of the truck' sense...

sumpfork: Right, I tested an corrected that note in the README. I don't think I'm comfortable pointing to non-official copies (though I don't even know if the fontsgeek ones really are legitimate). It's not so onerous to do this one-time conversion, and the fallback to Times honestly looks good enough for many people to not notice.

nickv2002: sure thing, just wanted to clarify

Issue #134, opened 2016-12-16
DividerDrawer.add_inline_images issue with new package format


wvoigt:
@sumpfork 
In Make better package (#132), an error was introduced.

```
replace = '<img src=' "'{}'" ' width={} height=' "'100%'" ' valign=' "'middle'" '/>'
replace = replace.format(DividerDrawer.get_image_filepath('coin_small_1.png'), fontsize * 1.2)
text = re.sub('(\d+)\s(c|C)oin(s)?', replace, text)
```

before this was:

```
path = os.path.join(self.options.data_path, 'images')
. . .
replace = '<img src=' "'%s/coin_small_\\1.png'" ' width=%d height=' "'100%%'" ' valign=' "'middle'" '/>'
replace = replace % (path, fontsize * 1.2)
text = re.sub('(\d+)\s(c|C)oin(s)?', replace, text)
```

The new code always picks `coin_small_1.png`.  The original code picked the `coin_small_x.png` where x was the number matched by (\d+).

Similar problem for the large coin and in debt section.

Other (dumb, newbie) questions:
1. Is there a reason it is called as `DividerDrawer.get_image_filepath` instead of `self.get_image_filepath`?
2. From a development standpoint, how are you running the program locally now.  I really don't want to have to install the package every time I make a change.  I prefer the "make a change an run" approach.  Is there a way to run this without installing it as a package?  I've been trying a local script in my own 'mytest' directory (at the same level as 'domdiv') called `dominion_dividers.py` that is simply:

```
import sys
import os

script_path = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))

if script_path not in sys.path:
    sys.path.insert(0, script_path)
    
os.chdir(script_path)

from domdiv import main

options = main.parse_opts()
main.generate(options)
```
It works for --help, but when generating an actual file, I'm getting an error (with an image filename) which brought me to the above problem.  But I don't think it is related.

sumpfork:
Use `python setup.py develop` and the `dominion_dividers` script it installs to run locally - it puts your dev version of the code on the python path, so you can test as you code.

I'll address the other stuff in a bit.

sumpfork:
Ok, #135 should address the main issue here, thanks for spotting that. The solution did require a bit more of a rewrite.

As for `DividerDrawer.get_image_filepath` - I like to do this for static methods to signal that that's what's happening (no `self` is actually passed in). Not a big deal to use `self` instead, though

[...]

PR #135, opened 2016-12-16
rewrite inline image replacement

sumpfork:
This addresses #134 - I had to rewrite the logic for image tag replacement because now that the images are in a package the code can't do path manipulations after the filename is looked up in the package. Instead, I now explicitly loop through all the tag matches in the card text and do the proper substitutions one by one.

I also took the opportunity to extract all of the actual replacement code so we can just loop through a set of replacement specs - these could be moved out into a datafile now.


wvoigt:
This works for me.
I tweaked the VP sizes a little to better match the coins.  Not sure why that changed, you used the same numbers from before.

PR #137, opened 2016-12-20
Fix CardSorter use of Base cards

wvoigt:


wvoigt:
@sumpfork This fixes 2 items:
1.  The order of the base cards depended upon the order in the database.  That used to be OK, but now we sort the database by name.  This adds an order to the base cards during sorting.
2. In cases where the language did not use "Base" as the base name, the base cards were not included.  Changed the code to look at the cardset_tag instead of cardset.

PR #140, opened 2017-01-08
make the translation file a markdown file

sumpfork:
Also blindly fix the language init script - looks like it may not be current anyway?


wvoigt:
@sumpfork  This looks good.   I added a plug for the translations on the main README.md.  I figured the more we can get to help on translations, the better.

PR #148, opened 2017-02-10
animals user expansion from BBG user rillaith

sumpfork:
BGG User rillaith sent me this via geekmail and I wanted to preserve it here.

Her comment:
```
I've been looking at the Dominion fan expansions, and working out how many more pages I need to print to update my current storage with the newer expansions etc, and as these things ever do, wandered through the rabbit hole of the internet on my way.

I saw a request from someone for sleeves for the fan expansions, with a response requesting help in the form of data entry & formatting. I had a bit of spare time, and figured I could do something vaguely useful, in scribing one of these for testing purposes as requested. I don't use Github, I'm afraid, so I wasn't able to branch the files and update them as requested, so I'm sending this via GM.

I picked the smallest (and am restraining myself from commenting on the quality!!), so have done Animals in both English & German. It's arguably the least useful based on its stats, sorry! But I figure it will give you some data to play with for categorising fan sets.

I've not updated the entirety of each file, but am just including the additional content. I've marked the start of each file with the filename.
```

I feel like we should handle fan expansions differently - maybe a separate list and option from the official ones? What you you think, @wvoigt and @nickv2002?


wvoigt:
I believe all the changes are in for full fan expansion support, using animals as the first entry.
Reviews and comments welcome.

domdiv/main.py
```diff
@@ -812,37 +835,105 @@ def filter_sort_cards(cards, options):
...
+        # Now keep only the cards that are in the expansions requested
+        cards = []
+        for c in Official_cards:
+            if c.cardset_tag in wantedSets:
+                cards.append(c)
+
+    # Take care of fan expansions.  Fan expansions must be explicitly named to be added.
+    # If no --fan is given, then no fan cards are added.
+    # Fan expansion names can be the names from the language or the cardset_tag
+    if options.fan:
```
sumpfork: Maybe we should provide an special value for --expansion, like `none', to turn off official expansions? I can imagine people printing fan expansions not wanting to also generate all the official ones, and I think there's no good way of turning them off other than producing an error via a bogus expansion selection right now?

domdiv/main.py
```diff
@@ -812,37 +835,105 @@ def filter_sort_cards(cards, options):
...
+        # First get the Fan set infomation
+        Fan_sets = set()  # Will hold fan sets
```
sumpfork: Can all of this code be shared with the version for the official expansions? In that case we can add the `none` keyword to both of them and default the fan expansions to use it unless otherwise specified.

wvoigt: Thinking out loud here. What makes it harder is that the absence of `--expansions` gives all cards where as the absence of `--fan` does not give any fan expansions. 

I just did a quick experiment.  It is possible to distinguish between the absence of `--expansions` (which returns None) and giving an empty `--expansions` (which gives in effect an empty list).

So another option instead of the `none` keyword might be to alias the absence of `--expansion` to be `--expansions *`.  Then giving `--expansion` without any expansion names could be used to signify no Official cards are wanted.  Then in this part of the code, we just combine the wanted/matched sets from `--expansion` and `--fan` and process them the same.  We add them if the card is in either set.

Not sure how this works on the web side.  If someone deselects all Official expansions, what happens?  Is an `--expansions` given without any actual expansion names?  If so, the above proposal would work without any changes there.

wvoigt:
@sumpfork 
OK, I took a belt and suspenders approach.  
I implemented the empty `--expansions ` and `--fan ` as outline previously as well as the `--expansions none`  and `--fan none` keyword approach.   In each of these cases no expansions will be shown for that type (official or fan).

Then I aliased the absence of `--expansions` to `--expansions * ` and the absence of `--fan` to be an empty list (or `--fan none`, or `--fan` which ever why you want to think about it.)

And finally, I combined as much code I as could for now, basically adding cards that are wanted for any of the official/fan sets specified.

sumpfork:
I have little time as usual, but a quick glance without testing makes it seem like your changes are reasonable, so I'm landing them as the new baseline.

PR #152, opened 2017-03-01
CZ language files

wvoigt:
Adding the cz language files.  Most of these still need translation.  However the names have been updated from the recent CrossReference.csv

Also added tools/add_language.py  to make it easier to create new language files from CrossReference.csv information.


msykora:
Sorry I didn't realize there was more to translate :( But would it be possible to publish it at least with Czech card names? To be honest, I need them to order the cards alphabetically in my box by names as I know them, I don't mind having the rules in English, I only read the text on the cards anyway :)

Issue #149, opened 2017-02-12
Second edition Dominion cards showing up twice in pdf. 


cboursnell:
When the "Dominion 2nd Edition" expansion is selected as an expansion the 7 new cards that were added over the 1st edition are appearing twice in the pdf. Artisan, Bandit, Vassal etc. 
You don't get this if you pick "Dominion 1st Edition" and "Dominion 2nd Edition Upgrade" however.

wvoigt:
This seems to be an unintended by-product of using startswith() when matching expansion names.
We had intended it to allow, for example, 'Dominion' to match all the expansions that start with "Dominion".  A kind of shorthand to get them all.

In this case, "Dominion 2nd Edition" will also match "Dominion 2nd Edition Upgrade" since they both start with the characters "Dominion 2nd Edition".

Looks like we can either:

- not use the startswith(), and require exact expansion names
- rename "Dominion 2nd Edition Upgrade" to something that does not match in that way (e.g., "Dominion 1st to 2nd Edition Upgrade"
- or do nothing

sumpfork:
I vote for exact matches.

wvoigt:
While it would take more work, I like the renaming option.  I like the way the upgrade sorts with 1st Edition instead of 2nd edition.

sumpfork:
I guess my only hesitation is the non-obvious nature of the 'startswith'-based implementation. I feel we should make groupings more explicit rather than relying on implicit naming parts.

wvoigt:
I understand why it is nice to have the exact match for the website.  It certainly prevents unintended selections.  But the startswith implementation was originally added so that someone could select 'dominion' and get all of the Dominion related expansions.  This is nice to have on the command line.
So how about this compromise:
- Make the exact names  'dominion' and 'intrigue' expand out to all of their expansions.  That is not one of the normal selection items on the web (unless you wanted it to be.)
- Add the ability to use `*` in the names to match multiple (or even hard to spell) expansions.  For example, `dark*` can be used instead of `'dark ages'` (note how you have to use the quotes because of the space on the command line).  `*upgrade` will get just the upgrades.

These are pretty much for command line use.  But then we have it both ways.

sumpfork:
As said, as long as the intent to include more than just `dominion` when specifying `dominion` is made explicit (e.g. through `dominion*`) I'm perfectly happy. We can probably use [fnmatch](https://docs.python.org/2/library/fnmatch.html) for a pretty easy implementation.

[...]

PR #153, opened 2017-03-01
Fix for Issue #149

wvoigt:
- Expansions names are no longer checked with "startswith()".
- Added wildcards for expansion selection.


wvoigt:
I shouldn't work on this so late. I have a simpler and more straightforward version in mind.

PR #154, opened 2017-03-02
Issue149

wvoigt:
- Expansions names are no longer checked with "startswith()".
- Added wildcards for expansion selection.


wvoigt:
Since the --fan addition in Pull #148 effects the same code, I'm closing this pull request in favor of Pull #148 which solves this problem, plus adds the --fan expansions.

PR #167, opened 2017-11-06
standardize arguments to use hyphens

openam:
Some of the arguments used hyphens/dashes `-` and some use underscores `_`. I think that they should at least be consistent. [GNU](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html) recommends using hyphens/dashes.

This is a breaking change.


sumpfork:
Good idea - I don't think it should break the web generator as the hyphens have to be converted to underscore to be valid python names anyway, but I need to find time to test it.

PR #180, opened 2017-12-08
correct the count of Platinum cards

sumpfork:
From BGG user Aziden:

`The count for Platinum (in Base/Prosperity) is incorrect. There are 12 Platinum, not 10.`


nickv2002:
I can confirm that [the Platinum pile should contain 12 cards](http://wiki.dominionstrategy.com/index.php/Platinum).

PR #181, opened 2017-12-08
Fix for base_cards_with_expansion option behaves differently #11

wvoigt:
Fix issue recorded at https://github.com/sumpfork/bgtools_web/issues/11

PR #182, opened 2017-12-10
Fix for base cards printing with expansions

wvoigt:
Fix issue recorded at sumpfork/bgtools_web#11
This is a more simple fix than the one proposed in Pull Request #181.

A few things to note:

1. As it is right now, per line 964 of main.py

```
    if options.base_cards_with_expansion:
        cards = [card for card in cards if card.cardset_tag.lower() != 'base']
    else:
        cards = [card for card in cards
                 if not cardSorter.isBaseExpansionCard(card)]
```
If you want the base cards with the expansion, it will not print the base cards separately.  That may not be intuitive via the web interface.

2, As it is right now, when included in an expansion, the base cards are sorted to the end.

### DIFF
--- a/domdiv/main.py
+++ b/domdiv/main.py
@@ -961,9 +961,9 @@ def filter_sort_cards(cards, options):
     cardSorter = CardSorter(
         options.order,
         {card.card_tag: card.name for card in cards if 'base' in [set_name.lower() for set_name in card.cardset_tags]})
-    if options.base_cards_with_expansion:
-        cards = [card for card in cards if card.cardset_tag.lower() != 'base']
-    else:
+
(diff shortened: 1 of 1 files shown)


wvoigt:
I guess to put it another way, what should be the output if both `--base_cards_with_expansion` and `--expansions base` are given? 
I suspect that we want to do both.  Print the base cards together and also print the base cards with the expansion.   If that is the case, then we will want to remove line 965.

sumpfork:
I agree that if ‘base’ is specified in expansions, a base expansion set should always be printed. And if base cards are requested in the various expansions that have them in addition, they should appear.

wvoigt:
@sumpfork 
I think this is ready.

PR #187, opened 2018-01-06
Dutch translation for Empires - Keizerrijken

martdegraaf:
Translation for Empires in Dutch.

Only description of base cards.

I just bought 'Empires' in Dutch and that is the only expansion i have
that is not translated, so thats why i contributed. :)

+ some spelling errors fixed in other sets such as
++ langdoedkaarten
++ Onderdakkaarten (shelters)


martdegraaf:
Could you pull this in? Then i can print my Dutch Empirers tabs 👍  :)

sumpfork:
Hi, could you merge in master to check the unit tests? I updated the tests to at least load all language files, and some files here appear to be invalid JSON. I should really also add a test that makes sure the language tool has been run in addition, but for now at least being able to run the script would be good :).

Thanks!

martdegraaf:
Hi Sumpfork,
I edited it now in an IDE, the tests seem to work now, Thank you for providing this package!

sumpfork:
Landed with 2 final fixes as #189. Thanks a lot!

PR #192, opened 2018-01-21
Test language tool run

sumpfork:
This is a minimal integration of @wvoigt's language fix up tool into the unit tests. Mainly makes sure that if we run the language tool and diff the output to what's checked into the `domdiv` package there are no important differences - especially, nothing got reverted to English because the 'untranslated' field was not updated. This is the main mistake I've seen from helpful contributions providing translations.

This PR also
- minimally turns the language tool into a more easily runnable script by taking arguments and not assuming that it's run from a specific directory
- switches testing to py.test - it seems to be the thing nowadays, and I have more experience with it than other frameworks

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ python:
 # command to install dependencies
 install:
   - pip install -U setuptools
+  - pip install -U pycodestyle
   - pip install .
 # command to run tests
 script: py.test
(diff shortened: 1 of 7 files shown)


sumpfork:
Note: github doesn't do a great job on the diffs for the tests. The main thing that changed was that I took out mentions of `unittest` and took them out of classes for now - we aren't really using anything about the classes so far.

Issue #184, opened 2017-12-16
Expansion Dividers card display and card count


wvoigt:
It used to be that the number of unique cards in an expansion was the same as the number of Kingdom Cards (and hence Randomizers) for that set.  That is no longer the case.  Some sets now have several non-Kingdom Cards that do not have Randominzers.  The card counts for the Expansion Dividers should match the number of Randomizers (for correct wrapper size, as well as indicating the number of Randomizers.)  The code right now uses as the count the number of cards listed in the Expansion. 

In addition, ideally the card list on each expansion divider should indicate which cards are Kingdom cards and which cards are not Kingdom Cards.

wvoigt:
This should be fixed with Pull #196 
On the expansion dividers, this fix makes the card count is the number of randomizer cards for that expansion.  On the alphabetical list of cards printed on the expansion divider, regular text font is used for the cards with randomizers (as in the past), and cards without randomizers are displayed in italics.  This works with or without card groupings.

In addition to helping keep track of randomizers, this also makes sure the expansion slip covers are the right size to hold the randomizer cards.

sumpfork:
Done with #196, release pending.

PR #195, opened 2018-03-15
Randomizer Fix

wvoigt:
This fixes Issue #184 
On the expansion dividers, the card count is now the number of randomizer cards for that expansion.
On the alphabetical list of cards printed on the expansion divider, regular text font is used for the cards with randomizers (as in the past), and cards without randomizers are displayed in _italics_.

In addition to helping keep track of randomizers, this also makes sure the expansion slip covers are the right size to hold the randomizers.


wvoigt:
I'm having trouble with the new CI checks.  It does not like the new
` "randomizer": false, `
entries in the `cards_db.json` file.

sumpfork:
You should be able to run it locally using `py.test` for easier checking.

What that test does is run your language update script, then run a diff on what it generates and what is in the `card_db` dir in `domdiv` to make sure there are no differences. Looks like maybe the lines are being reshuffled? Just run the language update tool first and check its output in if it looks ok.

wvoigt:
@sumpfork 
The only thing that changed was adding 
`"randomizer": false, `
to some entries in the `cards_db.json` file.  
Nothing that should have changed anything with the languages files.  And I didn't reorder any cards in the db.  Interesting.

sumpfork:
I'm saying the update language script is probably reordering the fields from the order you checked in. Just check in what it generates.

PR #196, opened 2018-03-16
Expansion Randomizer Count Fix

wvoigt:
This fixes Issue #184 
On the expansion dividers, the card count is now the number of randomizer cards for that expansion.
On the alphabetical list of cards printed on the expansion divider, regular text font is used for the cards with randomizers (as in the past), and cards without randomizers are displayed in italics.

In addition to helping keep track of randomizers, this also makes sure the expansion slip covers are the right size to hold the randomizers.


wvoigt:
@sumpfork 
This is a cleaner attempt at the fix.  I'm still not sure why it insists every line of the `cards_db.json` file has changed.  I only added 25ish "randomizer" properties.

wvoigt:
@sumpfork 
Finally got the line endings to work out.

domdiv/cards.py
```diff
@@ -56,6 +56,17 @@ def __init__(self, name=None, cardset='', types=None, cost='', description='',
             self.count = []
         else:
             self.count = [int(count)]
+        self.randomizer = randomizer
+        if self.randomizer:
+            # Remove classes of cards that never have Randomizers
+            # Card types
+            if any(self.isType(t) for t in [
```
sumpfork: Is there a strong reason to not just set these individually to `false` in the db files? I'd like to try to get rid of as many places as possible where cards receive special handling by type/name in addition to what's in the DB.

If there is a strong reason, I think we should at least make a new module (`constants.py`) that collects all of these lists of card names as well as other constants currently in use like the various option lists.

domdiv/main.py
```diff
@@ -971,17 +972,36 @@ def filter_sort_cards(cards, options):
     if options.expansion_dividers:
 
         cardnamesByExpansion = {}
+        randomizersByExpansion = {}
```
sumpfork: There's a class called `Counter` in the standard library `collections` module that would clean up this code a bit (get rid of the `setdefaults` below etc.) Also, this should maybe be called `randomizerCountsByExpansion`.

Similarly, there's `defaultdict` in the same module that would let you define `cardnamesByExpansion` as `defaultdict(dict)` and avoid having to explicitly pre-fill new entries below.

domdiv/main.py
```diff
@@ -990,12 +1010,22 @@ def filter_sort_cards(cards, options):
                     if 'short_name' in set_values:
                         exp_name = set_values['short_name']
 
+                card_names = []
+                for key, n in sorted(cardnamesByExpansion[exp].items(), key=lambda (k, (x)): x['sort']):
```
sumpfork: You don't need any parens around the `lambda` args.

domdiv/main.py
```diff
@@ -971,17 +972,36 @@ def filter_sort_cards(cards, options):
...
+                order = 0
+                if c.card_tag in cardSorter.baseOrder:
+                    # Use the base card ordering
+                    order = 100 + cardSorter.baseOrder.index(c.card_tag)
+                cardnamesByExpansion[c.cardset][c.card_tag] = {'name': c.name.strip().replace(' ', '&nbsp;'),
+                                                               'randomizer': c.randomizer,
+                                                               'count': 1,
+                                                               'sort': "%03d%s" % (order, c.name.strip(),)}
+            else:
+                cardnamesByExpansion[c.cardset][c.card_tag]['count'] += 1
```
sumpfork: Maybe I don't totally understand the intent here - how is this count ever > 1? Don't we always have a single instance of each card name per expansion? At least when I run this this is the case - or is it in combination with some other option that I'm not thinking about?

domdiv/main.py
```diff
@@ -971,17 +972,36 @@ def filter_sort_cards(cards, options):
...
 
         for set_tag, set_values in Card.sets.iteritems():
             exp = set_values["set_name"]
             if exp in cardnamesByExpansion:
                 exp_name = exp
 
-                count = len(cardnamesByExpansion[exp])
+                count = randomizersByExpansion[exp]
```
sumpfork: I'm not totally sure making the count here the randomizer count is intuitive without explanation. Maybe the count should stay the same as before, but the randomizer count get added somewhere else (end of the card list?) with some explanatory label?

domdiv/cards.py
wvoigt: I agree.  I'll move this into the db files.

domdiv/main.py
wvoigt: Nice!  I'll make the change to use these.

domdiv/main.py
wvoigt: I'll fix that.  It was a hold over from a point in time when I was having issues with the lambda function.

domdiv/main.py
wvoigt: Ahhh, well I cheated.  It can be greater than one if you break up Curses or add starting decks.

domdiv/main.py
wvoigt: First, this is only for the expansion dividers.  Regular card dividers naturally have card counts that match the number of cards for that divider.  

But what is the best for the expansion dividers?  Now I admit I have more of a slip cover view, which is what I prefer to use.  So what does "count" mean for an expansion divider?  Currently, the count depends upon which groupings are selected.  You get one count if you group specials, another if you don't, and another if you move Events and Landmarks at the global level, etc.  The count varies.  None of these can really be used for slip covers - since there really are no cards of those types to hold in the slip cover.

What the slip covers *could* be used for is storing the randomizers for the expansion.   So at a minimum, I would propose using the randomizer as the number of cards used to generate the depth for the slip covers.   And right now, for all the non-expansion dividers in draw.py, that is the card count.

In the end, due to the variance, I thought it would be better to use the number of randomizers as the card count, and then just differentiate the cards in the card list by randomizer (Kingdom cards) vs non-supply cards.  So the count icon to quickly see if you have the right number of randomizers, and the list to see if you have all the right cards dividers.

sumpfork: I see, ok. I don't have a problem with putting the number of randomizers, I just feel we're going to get bug reports saying "the number is wrong" if we don't explain it on the divider somehow. On the other hand, it's an optional (but nice) feature anyway, so maybe it's fine without explanation. Certainly not worth holding up the code over.

wvoigt:
@sumpfork 
Interesting CI fails.  It is now flagging all backslash sequences.  I did not get this with my local run.  Did something change?  Or do they need to be coded differently?

wvoigt:
@sumpfork, I've fixed the CI issues.  Please take another look.  Thanks.

Issue #175, opened 2017-11-25
Additional Divider Options: Curse, Trash, Starting Deck


Zannick:
My current organization uses a Curse divider for 10 cards apiece so I don't have to do any counting at setup, so it would be great if there were an option to add more Curse dividers to the pdf (as opposed to reprinting a page).

I also keep Starting Decks separately, and thanks to having Dominion/Intrigue I can have up to 8 of these.

Additionally, I don't care about the Trash dividers that show up for 1st Edition Dominion/Intrigue.

In short, I'm suggesting a "Starting Deck" base divider plus the ability to tweak the numbers of base dividers.

wvoigt:
@Zannick 
I am working on options for `--no-trash`, `--curse10`, `--start-decks`.
I need to figure out how to represent the start deck dividers and the extra curse dividers on the expansion dividers.  It looks a little odd alphabetically with the extra entries.

sumpfork:
Done with #201, release pending.

PR #197, opened 2018-03-24
Additional Divider Options: Curse, Trash, Start Decks

wvoigt:
This provides the features requested in Issue #175 
It relies on changes made for the randomizer card output on the expansion dividers (Pull #196 )

### DIFF
--- a/domdiv/main.py
+++ b/domdiv/main.py
@@ -292,6 +292,21 @@ def parse_opts(cmdline_args=None):
         help="Group cards that generally are used together "
         "(e.g., Shelters, Tournament and Prizes, Urchin/Mercenary, etc.).")
     group_select.add_argument(
+        "--no-trash",
+        action="store_true",
+        dest="no_trash",
+        help="Exclude Trash from cards.")
+    group_select.add_argument(
+        "--curse10",
+        action="store_true",
+        dest="curse10",
+        help="Package Curse cards into groups of ten cards.")
+    group_select.add_argument(
+        "--start-decks",
+        action="store_true",
+        dest="start_decks",
+        help="Include four start decks with the Base cards.")
+    group_select.add_argument(
         "--include-blanks",
         action="store_true",
         help="Include a few dividers with extra text.")
@@ -525,6 +540,14 @@ def get_resource_stream(path):
     return codecs.EncodedFile(pkg_resources.resource_stream('domdiv', path), "utf-8")
 
 
+def find_index_in_list(lst, key, value):
+    # Returns the index of the first dictonary in lst that has key with value.  Otherwise returns None
+    for i, d in enumerate(lst):
+        if getattr(d, key, None) == value:
+            return i
+    return None
+
+
 def read_card_data(options):
 
     # Read in the card types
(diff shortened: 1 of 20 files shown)


domdiv/main.py
```diff
@@ -525,6 +540,14 @@ def get_resource_stream(path):
     return codecs.EncodedFile(pkg_resources.resource_stream('domdiv', path), "utf-8")
 
 
+def find_index_in_list(lst, key, value):
+    # Returns the index of the first dictonary in lst that has key with value.  Otherwise returns None
+    for i, d in enumerate(lst):
```
sumpfork: This appears to not be looking a dictionaries, but at objects' attributes, contrary to the comment. It should also have a more specific name that reflects this, like `find_attribute_value` or similar. 

This could also be written inline or implemented as as `next((i for i, o in enumerate(lst) if getattr(o, key, None) == value), None)` though I'm not sure that's clearer.

domdiv/main.py
```diff
@@ -560,6 +583,57 @@ def read_card_data(options):
...
+            # Add the new dividers
+            cards.extend(new_cards)
+
+    # Create Start Deck dividers. 4 sets. Adjust totals for other cards, too.
+    # The card database contains one prototype divider that needs to be either duplicated or deleted.
+    # Do early before propagating to various sets.
+    i = find_index_in_list(cards, 'card_tag', 'Start Deck')
+    if i is not None:
+        new_cards = []
+        if options.start_decks:
```
sumpfork: For clarity, can you move this `if` to the outer level and change the logic about deleting the start deck entry below to us a flag?

Maybe also print a warning if the option is specified but for some reason the `Start Deck` db entry isn't there.

domdiv/main.py
```diff
@@ -560,6 +583,57 @@ def read_card_data(options):
...
+            # Get the index for Copper and Estate cards so their totals can be adjusted
+            c_i = find_index_in_list(cards, 'card_tag', 'Copper')
+            e_i = find_index_in_list(cards, 'card_tag', 'Estate')
+            if c_i is not None and e_i is not None:
+                # Take card counts out of Copper and Estate for 4 Start Decks
+                cards[c_i].setCardCount(cards[c_i].getCardCount() - (4 * 7))
```
sumpfork: Can we move `7` and `3` to be defined constants here and below?

Same for the `4` - in fact, maybe this should be specifiable using an option that defaults to 4, considering that someone might easily want to store only 2 or maybe as many as 5 or 6 starting decks if they play with those numbers of players often.

domdiv/main.py
```diff
@@ -560,6 +583,57 @@ def read_card_data(options):
...
+                cards[e_i].setCardCount(cards[e_i].getCardCount() - (4 * 3))
+                # Add cards to Start Deck prototype
+                cards[i].setCardCount(7)
+                cards[i].addCardCount([int(3)])
+                for x in range(0, 3):
```
sumpfork: See above, should depend on number of starting decks specified.

wvoigt: I actually started to make the number of starting decks an option, but then stopped after I thought about it a bit. 

1 to 4 is easily done.  But remember that this applies to all the base cards.  So let's say the user picks 2.  So it makes 2 start decks for Dominion 1st edition,  2 start decks for Dominion 2nd edition, 2 decks for Intrigue 1st edition, (none for 2nd addition), and 2 start decks for the Base.  So you really don't get 2, you get 2 x the number of expansions that you include that contain base cards.  So two if not really 2.

The harder case is picking something greater than 4.  After 4, a given set does not have enough Copper and Estate cards.  So where do those come from?  From a slip case point of view, we need to keep the counts so the slip cases are the right depth.  And we never really know which expansions the user will include.   I know it could be done, but even more options/information would be needed from the user to do it "right" for them (which might be different from others.)
The same problem occurs if you really want "2" to mean "2" above.  From which expansion do we make these starting decks from?

In the end, I settled on the fact that the game is designed for 1 to 4 players each with starting decks.  The number of Estates used in a 3 and 4 player game (12) matches exactly the number that is left over after you make 4 starting decks.  No special counting.   While the total number of starting decks is 4x the number of expansions with base cards, in the end,  I figured if they want starting decks they will most likely be consistent and want them across the board.  And after a while, these cards are extras anyway and it keeps them from having to remove the extra Estates.

sumpfork: Yeah, ok, you're probably right. Let's leave it as you have it.

domdiv/main.py
sumpfork: Sure, `find_index_of_object` sounds good.

domdiv/main.py
wvoigt: So I thought about this some more.  And maybe there is a middle ground.

If we limit the starting decks to only the Base (and not included in Dominion, et. al. with the `--base-cards-with-expansion`) and then make the option `--start-decks` accept 0 - 4, I think this may work. 

If someone wants 6 starting decks, they have to run the program twice, once with 4 and once with 2.  I guess they could do 3 and 3, but the Estate counts will be weird (but correct!).  In any case, for that printed Base set, the card counts would be correct.

The only oddity is if they are trying to make base starting decks and they don't have the cards from the other sets.  They would just have to substitute the Start Decks, Copper, and Estate dividers from the Base into their Dominion set - it would not be printed out exactly as used with the `--base-cards-with-expansion`.

wvoigt: For now, I left the above option out and have keep this to doing 4 start decks.  But I did put in some better infrastructure if we do want to change to a variable number of start decks.  The biggest change was allowing for any number of attributes to be matched in `find_index_of_object`

If we do change:
* the location this code would need to change to AFTER the cards are expanded out for all the sets
* The card index searches with `find_index_of_object` would need to add a 'cardset': 'base' attribute
* and the argument would need to take an integer (default to 0) which would be used in the copy loop.
Then everything is localised to only the base set.  Not to any base cards added to other sets.

wvoigt: For some reason, I'm still getting CI failures.   It passes on my local system.  Has anything changed?

All are "W605 invalid escape sequence" in draw.py and cards.py

wvoigt:
For some reason, I'm still getting CI failures. It passes on my local system. Has anything changed?

All are "W605 invalid escape sequence" in draw.py and cards.py

wvoigt:
@sumpfork, I've fixed the CI issues (and updated my test system as well).   Please take another look and let me know if you have other comments.

sumpfork:
I think you need to do a PR against current master for me to land this.

wvoigt:
A cleaner version of these changes is in Pull #201.

Issue #120, opened 2016-11-25
Add a command option to include a page in the pdf with the options used


wvoigt:
Since it can be months between printing/generating dividers, it would be nice to have a way to help remember the options used to get a particular output.

Presumably as the last page.
Hopefully formatted nicely.
Possibly with a list of all the options possible (i.e., --help)

wvoigt:
First try at this is in Pull #150

wvoigt:
Instead of making this a part of the draw refactor, I just put the changes for this issue into a separate Pull request #198

sumpfork:
Sure, if you have the time. I'm off to Germany for the next two weeks, so there's a chance that between visiting and taking care of kids I'll have a chance to finally get to land some of your PRs :).

sumpfork:
Done with #198, release pending.

PR #198, opened 2018-03-24
Add option to add pdf page with option information

wvoigt:
This is an attempt at Issue #120.
This adds options --info and -info-all:
--info to add a page that has all the options used for the PDF file.
--info-all is the same as --info, but includes pages with all the possible options that can be used.


sumpfork:
Seems to work, though the long output is harder to read than on the command line because the text after each option isn't indented. Is it for you? Otherwise, no big deal, it's a good start.

wvoigt:
Let me see if I can fix that.  My intent was to have some indentation.

wvoigt:
@sumpfork The indent problem should be fixed.  It is much easier to read now.

Issue #204, opened 2018-07-13
Clean up image file access code to only use no-coin image names


joeskeen:
There are a handful of images referenced in `types_db.json` that do not exist in the repository:

* duration-reaction.png
* reserve.png
* reserve-victory.png
* action-treasure.png
* boon.png
* event.png
* expansion.png
* hex.png
* landmark.png
* night.png
* state.png
* reserve-treasure.png

In each of these cases, it appears that there is a corresponding `{CARD_TYPE}_nc.png`.  Would it be best to change all images in the file to use the `_nc` suffix, or would it be better to generate the missing images?

Either way, I would be happy to submit a PR.

joeskeen:
I noticed that currently the `_nc` is being hardcoded into the image file name (see https://github.com/sumpfork/dominiontabs/blob/236feff33c2320995a0c7935503bf35c07a76862/domdiv/cards.py#L233).  Would it be appropriate to change the `types_db.json` to reference the correct image file and remove the path splicing?

sumpfork:
We switched to using no-coin images and generating/placing the coin image separately quite a while ago, so there's no need to have coin images. You are right that it's confusing that the images referenced in the db files do not (and should not) actually exist. I'll reword this task to ask for a cleanup of that.

sumpfork:
So here's what should probably happen:
1) get rid of the old `getNoCoinTabImageFile()` and change all occurrences to reference `getTabImageFile()` instead
2) rename all relevant image files to get rid of the `_nc` part, getting rid of existing files with that name as far as they exist.

Note that this is all just code readability maintenance - there's no visible bug, as only `_nc` images are referenced, and they all exist.

PR #205, opened 2018-07-13
make no-coin images the default, clean up code and file names

sumpfork:
Refs #204


joeskeen:
👍 It doesn't look like I can review this but it looks good to me according to my investigation in #204 :)

PR #206, opened 2018-07-15
exclude start deck from randomizer

joeskeen:
As the start deck will always be there, IMHO it should not be included in a randomizer (there is no Start Deck randomizer card in Dominion sets).


wvoigt:
I agree this flag makes sense and is consistent with the other base cards.  At the moment it does not change any output, since the base cards (including the start decks) are handled a different way.

PR #208, opened 2018-07-24
Python 3 Compatibility

sumpfork:
This makes the code python3 compatible, retaining python2 compatibility.

Also simplifies the unit tests a bit, and makes them run on python 3 on Travis CI.

Do note that (I think because of a newer reportlab version) I had to close all the `<para ...>` tags our code introduces - there may be formatting differences that I didn't catch.


domdiv/draw.py
```diff
@@ -922,6 +930,7 @@ def drawText(self, card, divider_text="card", wrapper="no"):
                     dmod = d
                 else:
                     dmod = self.add_inline_images(d, s.fontSize)
+                print('PARA ({}): "{}"'.format(card.name, dmod))
```
nickv2002: Not sure what this new print statement is doing. Not saying it shouldn't be here it just isn't replacing anything obvious in the diff and warrants a double check.

sumpfork:
@nickv2002 you can see a no-whitespace changes diff by appending `?w=1` to the PR url, e.g. https://github.com/sumpfork/dominiontabs/pull/208/files?w=1

nickv2002:
Nice: I don't use GitHub much so I didn't know about that flag for whitespace. Looks like it's an in the **Diff Settings** dropdown too. Thanks & great work here.

PR #209, opened 2018-07-30
decode files as utf-8 on load

sumpfork:
One more - this makes it work under python 3.5 (higher versions auto-decode these inside the json package.)


nickv2002:
Seems reasonable. 👍

PR #217, opened 2018-10-08
Black avery background

sumpfork:
This adds an option `--black-tabs` that makes the background of the tab portion of the dividers black.

In the process, changes all the tab images to have proper transparency.


wvoigt:
So I gave this a go.  In general the tabs look like you would expect.
Now I didn't print on actual labels.  I just looked at the output.

My one comment is that this will require very precise lable and printer alignment assuming they are printed on standard white labels.  Any straying, and the tabs will have slivers of white.   
Would it make sense to have a extra black around the tabs, just in case they are not perfectly aligned?  Or even just have the whole background printed black.  I know that wastes more ink, but it would guarentee no partially white labels.

Issue #213, opened 2018-09-04
German text on Adentures


idlaviV:
If I want to create Adventure-tabs with german text, I get "Server Error (500)" and there is no pdf. Also, there is no preview of the set in the right column.

The issue seams to lie at the card-texts of Adventures itself, if I choose "blank" or "rules" for both sides, the file is generated without an error. I could not reproduce the error for other sets than Adventures.

sumpfork:
Yes, the tags in other language need to be fixed - my mistake, as I didn't catch them all when I fixed them for English. There's also the issue that more card texts are used when grouping special cards, which is the default for the online generator.

I'll fix them and also add a test that catches this in the future.

idlaviV:
I saw that you commited new code and it works now. However, there seems to be an analogous error when I include expansion dividers (in english or german).

sumpfork:
I believe this is fixed in 3.5.1, which is now deployed. Please reopen if not.

PR #219, opened 2018-10-22
Fix special card groups option for non en_us languages

sumpfork:
Right now french and dutch throw exceptions if `--special-card-groups` is used, including on the web generator (#213 was the issue for the German version). This fixes some of the tag issues in the db, and adds a generation test that generates all languages with no options, and with the `special-card-groups` option. It also produces more meaningful exceptions for these cases, and fixes a warning for regexes in python 3.


sumpfork:
To note, I did little checking on how good the dividers look given these fixes for these languages - there should probably be a pass to make them look ok, but I don't have time for that.

PR #222, opened 2018-11-14
Added missing translation

GauGau2:
Only 3 missing translations left


GauGau2:
Tried to send by email, which apparently doesn't work. Trying to submit pull request for the second time. Hope this will work better than my last attempt.

Issue #223, opened 2018-11-14
Blank Tabs Print Option


nickv2002:
[As requested by David Jaehn on BGG](https://boardgamegeek.com/article/30470696#30470696):
> Is there any way to print out "blank" tabs (so I could at least pre-print them and label them pretty when you do have them done

I can also see this as being useful if you wanted to laser cut out the dividers after printing them. Not sure if they would be the same as what David was thinking, but for laser cutting, you would just want a completely blank divider with a hairline outline (showing the laser where to cut).

wvoigt:
Added int parameter to `--include-blanks`, which as far as I could tell was not used anymore.  Then updated main.py and draw.py so that the given number of blank dividers was added to the cards to be printed under the "Extras" set.

This and other changes are in Pull #224.

sumpfork:
I guess this should probably be added to the online generator before I close it.

Issue #200, opened 2018-04-10
UK Avery labels


jeremyscode:
Unfortunately it's almost impossible to get the 8867 labels in the UK for a reasonable price. Is there any chance you could add an option for the UK version? I don't mind testing it for you, I'd imagine the online template designer would provide enough detail for coding it in theoretically. I'll let you know if it works. Link here https://www.avery.co.uk/product/mini-multipurpose-labels-l4732rev-25

sumpfork:
This is landed, but needs to be integrated with the online generator.

PR #224, opened 2018-11-16
Add generic label and blank tabs support

wvoigt:
For issue #200:
Created a labels_db.json file to store label information.
Added command line parameter `--label` that takes one argument for the name of the label to use.
Values from that label then override other options as needed.  In addition, the hardcoded dimention overrides for `--tabs-only` are made generic based on the label specified.
I also allowed for a body size to be specified with the label.  This is for two reasons:

1. Some labels are really too tall to look right.  By specifying a small body size, then the card count and card type information can be added (if the options are specified).
2. In the future, when I get around to updating draw.py to allow for divider rotation, then we can more easily add support for larger labels (that usually will require a 270 rotation) that can be used with BCW Trading Card Dividers that are 2.68" x 3.81".  There are Avery labels that could fill most of that.  It would allow for full dividers without the need to cut (even if they are more expensive).

**Note:** I did my best on getting these label dimentions correct from the templates.  I'm pretty comfortable with the default 8867 since those numbers match what was hardcoded previously.
Someone who has access to the UK A4 labels would need to try these out to see if any margin tweaking was necessary.

For issue #223:
For the blank tabs, I added an int parameter to ` --include-blanks`, which as far as I could tell was not used anymore. Then updated `main.py` and `draw.py` so that the given number of blank dividers was added to the cards to be printed under the "Extras" set.

Please let me know if I missed anything or if something does not look right.


domdiv/main.py
```diff
@@ -498,6 +506,48 @@ def clean_opts(options):
...
+        with get_resource_stream(labels_db_filepath) as labelfile:
+            label_info = json.loads(labelfile.read().decode('utf-8'))
+        assert label_info, "Could not load label information from database"
+        for label in label_info:
+            if options.label_name.upper() in [n.upper() for n in label['names']]:
+                options.label = label
+                break
+
+        if options.label is None:
+            print("Error: Label not defined")
```
sumpfork: I think this should fail/throw an exception rather than subtly printing to stdout.

domdiv/main.py
```diff
@@ -892,6 +942,16 @@ def filter_sort_cards(cards, options):
 
         cards = keep_cards
 
+    # Add any blank cards
+    if options.include_blanks > 0:
+        for x in range(0, options.include_blanks):
+            c = Card(cardset='extras',
```
sumpfork: I'm getting
```
Traceback (most recent call last):
  File "/Users/pgorniak/.pyenv/versions/2.7.14/bin/dominion_dividers", line 11, in <module>
    load_entry_point('domdiv', 'console_scripts', 'dominion_dividers')()
  File "/Users/pgorniak/src/p/dominiontabs/domdiv/main.py", line 1372, in main
    generate(options)
  File "/Users/pgorniak/src/p/dominiontabs/domdiv/main.py", line 1347, in generate
    cards = filter_sort_cards(cards, options)
  File "/Users/pgorniak/src/p/dominiontabs/domdiv/main.py", line 1222, in filter_sort_cards
    cards.sort(key=cardSorter)
  File "/Users/pgorniak/src/p/dominiontabs/domdiv/main.py", line 822, in __call__
    return self.sort_key(card)
  File "/Users/pgorniak/src/p/dominiontabs/domdiv/main.py", line 808, in by_expansion_sort_key
    card.name), self.strip_accents(card.name)
  File "/Users/pgorniak/src/p/dominiontabs/domdiv/main.py", line 818, in strip_accents
    return ''.join(c for c in unicodedata.normalize('NFD', s)
TypeError: normalize() argument 2 must be unicode, not str
```
under python 2.

sumpfork: Also, maybe that means we should add a test for these options?

PR #226, opened 2018-11-20
Added translations for Nocturne and plurals

GauGau2:


GauGau2:
Not sure if the plurals are actually needed, please review technically

PR #227, opened 2018-11-20
Added renaissance translations for names

GauGau2:
As Renaissance hasn't been released yet in Germany the translation may be incomplte and inacurate


wvoigt:
It would make it easier if all your changes across multiple files were in one branch.  It would also then require only one pull request.

PR #228, opened 2018-11-20
Added partial translations from manual download section

GauGau2:
German Rio Grande sales partner Ass has released manuals with partial translations


GauGau2:
Yes, the sales partner's name is really "Ass" - it means "Ace" in German. Sorry for that.

Issue #235, opened 2018-11-26
Dismantle has no FAQ text (extra)


md5i:
All other supply cards have an FAQ (extra) entry that, usually, matches the official FAQ for that card on Dominion Strategy wiki.  Dismantle seems to be missing that.

nickv2002:
[Wiki FAQ is here.](http://wiki.dominionstrategy.com/index.php/Dismantle)

Issue #232, opened 2018-11-23
Missing types on Marauder


md5i:
Marauder is missing the Attack and Looter types.

md5i:
Also, Watchtower should include the Action type.

md5i:
I did a full search, and these seem to be the only two mismatches (for types).

Issue #231, opened 2018-11-23
Minor typos


md5i:
Minor typo in the text for Improve: "acrd" -> "card".

wvoigt:
Minor typo in the rules of Swamp Hag: "you will get..." -> "You will get..."

wvoigt:
Butcher should be reworded (I missed this in the other updates for issue #214.
"Take two Coin tokens. " should be "+2 Coffers (take two Coin tokens).<br>"

wvoigt:
I'm bunching these all up since it will take a run of /tools/update_language.py to fully incorporate these.  They can also be picked up the next time a language update is done.

PR #236, opened 2018-11-27
Text and database fixes

wvoigt:
Several text and database changes documented in Issue #235, Issue #232, and Issue #231.


md5i:
As long as this is still opern, Jack of All Trades should have "All" capitalized.

Issue #240, opened 2018-12-28
Option for Just Expansion Labels/Just Special Cards/Just Resources


Sunthief:
It would be nice to have an option to create a PDF with just the expansion labels, or just the resource/special cards (project etc.) labels in the form. You might need some a couple times.

ranneyd:
I had this exact same request.

I didn't fork or push anything because I'm lazy, but I cloned it and added this at line 1385 after `cards = filter_sort_cards(cards, options)`. Worked like a charm

```
    new_cards = []
    for card in cards:
        if card.isEvent() or card.isExpansion() or card.isLandmark() or "base" in card.cardset_tags or card.isType("Project"):
            new_cards.append(card)
    cards = new_cards
```

wvoigt:
Here is a generic proposal that I believe covers the above request and more.   Add two new options that limit/reduce the number of dividers to the matching type(s) given. This would happen after any other options are used to select expansions, group cards, etc. Basically this is done right before the cards are sorted before they are printed.

* `--only-type-any,` alias `--type-any`
keeps any dividers that have *any* of specified types

* `--only-type-all`, alias `--type-all`
keeps any dividers that match *all* of the specified types

For example:
`--only-type-any project expansion` would include in the output only expansion and project dividers
`--only-type-all duration attack` would include in the output only dividers that had both duration and attack types.
`--only-type-any landmark --only-type-all duration action` would include in the output any landmark dividers AND any dividers that had both duration and action.

The current list of types include:
Action, Artifact, Attack, Blank, Boon, Castle, Curse, Doom Duration, Event, Events, Expansion, Fate, Gathering, Heirloom, Hex, Landmark, Landmarks, Looter, Night, Prize, Prizes, Project, Reaction, Reserve, Ruins, Shelter, Shelters, Spirit, Start Deck, State, Trash, Traveller, Treasure, Victory, Zombie

Matches would be to these type names or their corresponding text in the language specified.  All case insensitive.  I have a prototype of this that I can post as a branch if this makes sense to everyone.

Comments?

sumpfork:
Deployed

PR #241, opened 2019-02-07
Add only-type-any and only-type-all options

wvoigt:
Inspired by Issue #240  I think this solves that issue as well as provides some added function.

This adds two new options that limit/reduce the number of dividers to the matching type(s) given. This would happen after any other options are used to select expansions, group cards, etc. Basically this is done right before the dividers are sorted just before they are printed.

* `--only-type-any,` alias `--type-any`, alias `--only-type`
  keeps any dividers that have _any_ of specified types
* `--only-type-all`, alias `--type-all`
  keeps any dividers that match _all_ of the specified types

For example:
`--only-type-any project expansion` would include in the output only expansion and project dividers
`--only-type-all duration attack` would include in the output only dividers that had both duration and attack types.
`--only-type-any landmark --only-type-all duration action` would include in the output any landmark dividers AND any dividers that had both duration and action.

Matches would be to these type names or their corresponding text in the language specified. All case insensitive. 

In addition, instead of using a hard coded list in main.py of all the supported types, code was added to automitically pull the supported types from a json file.   After implementing it, I thought a similar technique should be used for getting the supported expansions.  This way, a new expansion is automatically picked up as soon as the json set file is updated.

### DIFF
--- a/domdiv/main.py
+++ b/domdiv/main.py
@@ -27,12 +27,6 @@ LINE_CHOICES = ["line", "dot", "cropmarks", "dot-cropmarks"]
 
 EDITION_CHOICES = ["1", "2", "latest", "all"]
 
-EXPANSION_CHOICES = ["adventures", "alchemy", "base", "cornucopia", "dark ages",
-                     "dominion1stEdition", "dominion2ndEdition", "dominion2ndEditionUpgrade",
-                     "empires", "guilds", "hinterlands",
-                     "intrigue1stEdition", "intrigue2ndEdition", "intrigue2ndEditionUpgrade",
-                     "promo", "prosperity", "renaissance", "seaside", "nocturne"]
-FAN_CHOICES = ["animals"]
 ORDER_CHOICES = ["expansion", "global", "colour", "cost"]
 
 LANGUAGE_DEFAULT = 'en_us'  # the primary language used if a language's parts are missing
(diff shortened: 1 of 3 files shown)


domdiv/main.py
```diff
@@ -1359,6 +1430,54 @@ def filter_sort_cards(cards, options):
...
+        # Assemble a list of valid "all" types. Options are already lowercase.
+        for x in options.only_type_all:
+            if x in types_lookup:
+                type_known_all.append(types_lookup[x])
+            else:
+                type_unknown.append(x)
+
+        # Indicate if unknown types are given
+        if type_unknown:
+            print(("Error - unknown type(s): {}".format(", ".join(type_unknown))))
```
sumpfork: As usual, I think I'd rather have this quit with an error here.

domdiv/main.py
```diff
@@ -373,6 +407,29 @@ def parse_opts(cmdline_args=None):
         "--exclude-landmarks",
         action="store_true",
         help="Group all 'Landmark' cards across all expansions into one divider.")
+    group_select.add_argument(
+        "--only-type-any", "--only-type", "--type-any",
+        nargs="*",
+        action="append",
+        dest="only_type_any",
+        help="Limit dividers to only those with the specified types. "
+        "A divider is kept if ANY of the provided types are associated with the divider."
```
sumpfork: very minor, but missing a space at the end

domdiv/main.py
```diff
@@ -373,6 +407,29 @@ def parse_opts(cmdline_args=None):
...
+        "Default is all types are included. "
+        "Any type with a space in the name must be enclosed in double quotes. "
+        "Values are not case sensitive. "
+        "Choices available in all languages include: %s" %
```
sumpfork: please use `.format()` style string formatting.

domdiv/main.py
```diff
@@ -373,6 +407,29 @@ def parse_opts(cmdline_args=None):
...
+        ", ".join("%s" % x for x in TYPE_CHOICES))
+    group_select.add_argument(
+        "--only-type-all", "--type-all",
+        nargs="*",
+        action="append",
+        dest="only_type_all",
+        help="Limit dividers to only those with the specified types. "
+        "A divider is kept if ALL of the provided types are associated with the divider."
```
sumpfork: same as above, missing space

domdiv/main.py
```diff
@@ -373,6 +407,29 @@ def parse_opts(cmdline_args=None):
...
+        "Any type with a space in the name must be enclosed in double quotes. "
+        "Values are not case sensitive. "
+        "Choices available in all languages include: %s" %
```
sumpfork: `.format()`

domdiv/main.py
wvoigt: Done.  And I fixed a few others too in the arguments area.

domdiv/main.py
wvoigt: done

domdiv/main.py
wvoigt: Done

domdiv/main.py
wvoigt: done

domdiv/main.py
wvoigt: done

PR #243, opened 2019-02-15
Dutch translations for Empires expansion added

Emmavdv93:


sumpfork:
Thanks! The final step for updating language files is still a little obscure - I'm intending to fix that. See #244 for a version of this PR that passes the tests.

PR #256, opened 2019-06-01
add --exclude-expansion option (and test)

elrond79:
provides an easy way to include nearly all expansions, except for a select few

In my case, I noticed that you changed "--base-cards-with-expansion" so it doesn't exclude the "base" expansion automatically, and I wanted a way to do that.  Adding a new option seemed like a more generally useful way to go about it!

Also adds tests, both for this, and the "normal" --expansion option

Thanks again for the great tool!


elrond79:
closing this pr in favor of #257 (different branch name)

PR #259, opened 2019-06-08
add --tab-artwork-opacity flag

elrond79:
turn down background opacity to help text show up better against dark
backgrounds

this was particularly need on my printer, which I think prints a bit darker than normal

### DIFF
--- a/src/domdiv/draw.py
+++ b/src/domdiv/draw.py
@@ -6,11 +6,14 @@ import sys
 
 import pkg_resources
 
+from PIL import Image, ImageEnhance
+
 from reportlab.lib.units import cm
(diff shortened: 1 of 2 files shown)


src/domdiv/draw.py
```diff
@@ -1141,6 +1144,8 @@ def nameWidth(self, name, fontSize):
         return w
 
     def drawTab(self, item, wrapper="no", backside=False):
+        from io import BytesIO
```
sumpfork: Minor niggle, but let's move the import to the top of the file?

sumpfork:
Close enough.

PR #266, opened 2019-09-04
Added complete German translations for Renaissance

GauGau2:
Renaissance has been released two days ago in German by publisher Ass (sorry for the company name, they are really named that way). I got my copy on monday and started translating based on it, So I completed some names that were initially translated literally but have changed slightly upon release. I have added description and extras for all Renaissance extension cards.


sumpfork:
Replaced by #268

PR #268, opened 2019-09-16
Run language tool for #266

sumpfork:


sumpfork:
Replaces #266

beorn7:
Thank you very much for this. Would it be possible to tag a release with this included so that the German Renaissance cards are available on http://sandflea.org/?
Thanks again…

sumpfork:
Aber sicher doch, v3.8.1 ist jetzt online.

beorn7:
\o/

PR #278, opened 2019-11-17
German translation of Church promo card (Schweriner Dom).

daniel-schmidt:
Since I got this card today, I really would like to have a German divider for my collection :-)

The "description" text is the original one from the German card, the "extra" text is my own translation.


sumpfork:
Replaced by #291 which is now landed.

PR #281, opened 2019-12-02
Test french translation

wizloo:
Hello, I'm new to this but I want to contribute in french translation.
I've translated only one card yet : The Fortue Teller line 471. Tell me if I didn't make any mistake.
If it's good I'll do all the cards that I have ;)


sumpfork:
Replaced by #290 which is now landed.

PR #285, opened 2019-12-26
Update "Haunted Woods" FAQ/Extra text

cwhits:
Hi! Some text was missing from the FAQ for Haunted Woods. Just a couple of words at the beginning (the stuff before the colon).


cwhits:
I bet you can't guess which expansion I got for Christmas 😬

sumpfork:
Replaced by #289 which is now landed.

Issue #296, opened 2020-02-13
One card tab at the wrong position when using "--tab-number 2 --tab-side right-alternate"


Gnomiro:
I tested this on my local computer as well as on the website. If you generate the tabs with "--tab-number 2 --tab-side right-alternate" or in my case more specific:

dominion_dividers --papersize A4 --orientation vertical --back blank --front rules --count --tab-name-align edge --set-icon body-top --tab-artwork-opacity 0.8 --special-card-groups --expansion-dividers --centre-expansion-dividers --expansion-reset-tabs --exclude-events --exclude-landmarks --tabwidth 5 --tab-number 2 --tab-side right-alternate

The first card after the "Base" seperator is centered but only this one. If you use "--tab-side left-alternate" this is not the case. Maybe it is related to some modifications which are made for the Base-Card generation. Potentially the "tabOffsetBack" is not set back to 0?

In addition why are base cards not alternating their position? Is there any special reason for this?

wvoigt:
Fixed in pull #301

Issue #295, opened 2020-02-13
Expansion dividers with full tab width


Gnomiro:
Is it possible to make expansions dividers to have full width while having alternating not-full width tabs for other cards?

As fas as I searched in the code it seems that the tabWidth parameter is global and used for all cards, isn't it? It seems only to be possible to center make the expansion dividers centered with "--centre-expansion-dividers".

wvoigt:
Added this function in pull #301 
Added option `--full-expansion-dividers` that makes expansion dividers full width with centered text.

Issue #248, opened 2019-03-16
Avery 8867 verticle/horizontal offset option


timur-tabi:
Could you add an option to provide a user-specified offset of the entire page, vertically and horizontally?  The labels align perfectly now, which is great, but unfortunately for me, my plastic dividers are a little too short, and so I want the labels printed slightly higher on the page (closer to the top edge of the labels).

I was able to make the adjustment I needed by slicing 1mm off the top of the label sheet, then printing the page upside down.  But it would be nice if I didn't have to do that.

wvoigt:
This change is included in Pull #301

PR #301, opened 2020-03-18
Fixes for #284 #295 #296 #248

### LINKED ISSUE #284: Dot-cropmarks on back side are in wrong place
[sumpfork_dominion_tabs (3).pdf](https://github.com/sumpfork/dominiontabs/files/4002911/sumpfork_dominion_tabs.3.pdf)

Looks like for the back side, the offset is calculated from the opposite direction (right vs left)

wvoigt:
Various changes to fix issues #284 #295 #296.
Added `--full-expansion-dividers` option to force expansion tabs to be full width with centered text.

### DIFF
--- a/src/domdiv/draw.py
+++ b/src/domdiv/draw.py
@@ -129,6 +129,7 @@ class CardPlot(object):
         cropOnLeft=False,
         cropOnRight=False,
     ):
+
         self.card = card
         self.x = x  # x location of the lower left corner of the card on the page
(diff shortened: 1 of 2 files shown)


wvoigt:
Added fix for #248.  Added options `--front-offset` and `--front-offset-height` to give the ability to fine tune the front page (in a similar way the back pages can be tuned.)

wvoigt:
I think these are ready to go.  I batched up these small changes that help clear out some of the issue backlog.

Issue #273, opened 2019-10-17
Dominion Menagerie Expansion Support


nickv2002:
Here we go again. [Donald X says](https://www.reddit.com/r/dominion/comments/dg7lfd/errata_to_the_errata/f3ebqyq/):
> There will be an expansion next year, probably first quarter.

On one hand I'm excited but on another hand my current box with fancy dominiontabs dividers can **not** hold any more cards. 🙃

sumpfork:
Ha

nickv2002:
[Expansion announced](https://boardgamegeek.com/boardgameexpansion/298055/dominion-menagerie):

> Dominion, that's what you're trying to achieve. This time with animals! They each have a lesson to teach, whether it's how to spit really far or what kind of grass tastes the best. It's a lot to keep track of, but you're like an elephant: you remember everything. And you're afraid of mice. You've taken up riding. Horses are intimidating; they say you can lead a horse to water, but you haven't managed it, so you're working your way up, starting with dogs. So far so good; the dog hasn't bucked you off yet. Your menagerie got off to a poor start, with just a goat, two rats, and the advisor who suggested starting a menagerie. You couldn't get that fox you wanted, but it was probably bad anyway. Now you've got some camels, which are just as useless for sewing as you'd been warned, and a turtle that can hold its breath for longer than anyone can stay interested. Soon the animal kingdom will be yours.
> 
> **Dominion: Menagerie**, the 13th expansion to Dominion, has 400 cards, with thirty new Kingdom cards. There are Horses that save a draw for later, Exile mats that cards can be sent to and rescued from, and Ways that give Actions another option. Events return. You need the basic cards and rulebook in order to play Dominion: Menagerie, which can be combined with any other expansion in the game series.

nickv2002:
The [full Menagerie set set is available from the Dominion wiki](http://wiki.dominionstrategy.com/index.php/Menagerie_(expansion)
[Also rulebook PDF](http://riograndegames.com/getFile.php?id=2220)

nickv2002:
Not sure I'll find time to work on this soon given everything that's going on, but I'll poke at it when I do have time. Thanks and stay healthy everyone.

wvoigt:
I can start working on this.

PR #304, opened 2020-03-21
Menagerie changes

wvoigt:
Starting the Pull for menagerie so we have a place to discuss as items get added.
Reference Issue #273


wvoigt:
I'm seeing the conflict errors.   It has been a while since I did this and I know things have changed.  I'm assuming I need to change the files in card_db_src and not src/domdiv/card_db.  Is that correct? Or do I have it backwards?  And do I need to still add on the new language entries in English to all the other languages there?  Or is that done automatically now?

wvoigt:
FYI, to make this easier in the future (will there be any more expansions?), I'm entering the data in a spreadsheet (which is a lot easier than editing straight json.)  Then I will write a short script that takes a csv file and makes json code that can be added to `cards_db.json` and `en_us/cards_en_us.json` files.  I'll put them in then tools area when done in case anyone needs them in the future.

sumpfork:
Changes should only go into the top level db src dir, and ‘doit update_languages’ should run a modified version of your script to generate the version in the package. It’s still mostly similar to what you did, other than that it substitutes the English version wherever anything is missing, rather than requiring special markup as to what is untranslated.

wvoigt:
I could use some help proof reading.  I think I have most of this.  But a second set of eyes would be good.
[menagerie1.pdf](https://github.com/sumpfork/dominiontabs/files/4366177/menagerie1.pdf)
[menagerie2.pdf](https://github.com/sumpfork/dominiontabs/files/4366178/menagerie2.pdf)
[menagerie3.pdf](https://github.com/sumpfork/dominiontabs/files/4366179/menagerie3.pdf)

sumpfork:
Thanks, I'll try to get to this, and I apologize, there's something wrong with the travis check-in setup I tried to do to auto-compile the card db. I'll have to test it some more, but it seems to be getting confused about which branch to check out/commit to.

src/domdiv/tools/new_cards.py
```diff
@@ -0,0 +1,51 @@
...
+# coding=utf-8
+
+import csv
+import sys
+import json
+
+new_cards_db = []
+new_cards_en_us = {}
+
+with open(sys.argv[1]) as csvfile:
```
sumpfork: Let's use argparse, even for something this small?

wvoigt: > Looks like the tests need to be updated with the new card number?

Fixed

> I also have the concern that the non-JSON files with get out of date with the JSON files and be hard to reconcile. Are you thinking the CSVs should become source of truth? Otherwise there should probably be checks to make sure the scripts aren't duplicating cards, etc.

No, the JSON is still the source of truth.  This was really just a quick way to enter a bunch of new cards.  From past experience, it was probably 3x or more faster.  It allowed me to focus on the content and not deal with JSON formatting up front.   In the past, when editing straight JSON, there was a tendency to get lost in the trees.  And if something needed changed in several places, it was harder to find.
But once done, it was done.   In fact, I think I tweeked the JSON in a place or two.  So I don't view the csv as something that is kept up to date.  This is more of a one time use to make a bunch of edits easier and faster.

I really did not intend to spend too much time on it and make it too bullet proof.  The same problem of duplicate cards can happen when editing straight JSON (In fact, I have had that happen in the past when I cut-n-pasted JSON to enter cards.) I really only uploaded in case it would be helpful to someone in the future.

sumpfork: Ok, that's fine.

sumpfork:
https://travis-ci.org/github/sumpfork/dominiontabs/builds/666010019 - please install pre-commit. There's a note as to what to do in the README.

sumpfork:
And sorry travis isn't updating the PR statuses - not sure why, trying to fix.

Issue #247, opened 2019-03-14
Promo Expansion Symbol


TMDaines:
All the currently printed promo cards have the same symbol (the flag), even the older ones have now had a second printing with this symbol. Could the promo tabs be updated to all have the correct, current expansion symbol?

nickv2002:
My cards have the old symbol so I'd prefer those icons. (I think the original promo icons were pretty neat and expressive, soI miss the fact that they don't make them any more.)

So perhaps we could make it an option at some point. @TMDaines if you want to duplicate the result beforehand without much coding you could always just copy the flag icon image to have the same name as all the other icons and build your own. (I know it's not _easy_ though, because you still have to run the code locally instead of using the awesome web version.)

sumpfork:
I'd rather keep the unique symbols, too. If somebody wants to make a pull request to optionally standardize them to the new flag symbol, I'll be happy to accept it.

TMDaines:
If I could do this, I would, but I need to teach myself. Would love to have this option.

PR #307, opened 2020-03-24
Fix for #247 Promo Card Icons

wvoigt:
Added `--use-set-icon` option to force cards with unique icons to instead use their standard set icon.  This really only applies to Promo cards.

This is a fix for Issue #247


nickv2002:
@sumpfork have you added a checkbox on the web for the `--use-set-icon` CLI option? I couldn't find one, but would be something good to add to the extras section.

PR #308, opened 2020-03-25
Closed the unclosed tags

maksler:
Hi Peter,

I've found 5 html-tags that were still unclosed and fixed it.

Max


sumpfork:
I got interaction between travis and github working again so you can see remaining errors above. You can also reproduce them locally if you managed to run the `doit update_languages` via `python setup.py test`.

https://travis-ci.com/github/sumpfork/dominiontabs/builds/155235688#L761

maksler:
Hi Peter,

I'm clearly missing something obvious. When I run `python setup.py test` I get the same error as above, but even when I remove the entire line, it still gives me the same error.

I think I need to run doit update_languages, but that gives me the following error:

> ERROR: Two different tasks can't have a common target.'src/domdiv/card_db\' is a target for update_languages and update_languages.

maksler:
With the merge of PR316, I guess this also fixes the issues here.

As a beginner in Python, I'd still like to know why the checks kept failing there... @JSmid6, are you able to tell me?

JSmid6:
I'm not sure why yours was failing, for me it kept failing because the flake8 hook in pre-commit had been moved to a different repository

sumpfork:
> Hi Peter,
> 
> I'm clearly missing something obvious. When I run `python setup.py test` I get the same error as above, but even when I remove the entire line, it still gives me the same error.
> 
> I think I need to run doit update_languages, but that gives me the following error:
> 
> > ERROR: Two different tasks can't have a common target.'src/domdiv/card_db' is a target for update_languages and update_languages.

I'm really not sure what's going on here. What do `git status` and `git diff master` show for your local checkout?

Issue #166, opened 2017-11-05
Development Discussion: Should we change how groupings are designated?


wvoigt:
Some of this may have been covered by discussions in the past, but I could not find it.  A couple of things prompted this topic:

1. There seems to be a growing number of possibe multi-expansion cards (e.g., Events, Landmarks, Boons, Hexes, States, etc).  And right now, any grouping outside of the normal `--special_card_groups` requires code changes.
2. I noticed as I was entering Nocturne, that we have a number of "expansion Extras" sets that I don't think are used anymore.  Remnants of past implementations.

Also, I think it would be nice to see all the multi-expanion groupings handled in a uniform way, with the designations represented in the json instead of the code.  But before that, I think we should discuss what groupings should be implemented.

Right now, the possible groupings are:

- none - every card is printed out individually
- normal groupings - i.e., `--special_card_groups` Cards are grouped together and placed in the normal listing - usually alphabetical.  So this means that Landmarks are sorted along with all the other normal kingdom cards.
- global groupings - ie., `--exclude_events`, `--exclude_landmarks`, etc.  Cards are grouped at a global level all under an "Extras" divider.  A suggestion would be to simplify this to say `--global_card_groups`

What might be missing, is an option for groupings by expansion, with the cards placed in a "<expansion> Extras" divider.  This would keep them out of the alphabetical listing and potentially easier to find/gather for set-up.  A possible option name would be `--expansion_card_groups`.
So cards in `--expansion_card_groups` would be a subset of cards in  `--special_card_groups`, and cards in `--global_card_groups` would be a subset of cards in `--expansion_card_groups`.

So the question is, does this seem worthwhile?

And in any case, I think there should be a discussion on how to represent the global groupings in the json.  A brute force method might be to add properties such as
```
    "card_tag": "Miserable - Twice Miserable",
    "cardset_tags": [
        "nocturne"
    ],
    "cost": "",
    "count": "6",
    "group_tag": "nocturne states",
    "group_extra": true,   <-- New
    "group_global": true,  <-- New
    "types": [
        "State"
    ]    
```

The only issue I see is every card would need this when we really only need it once.  Another option would be to add it only to the card that has the `"group_top": true,` designation (which is used to designate the card that is used for cost and type information for the group).  Another option would be to create some other form of json structure to take care of groups.
Thoughts?

nickv2002:
[...]

PR #311, opened 2020-04-04
Fix issue#166

wvoigt:
I took a look at this yet for issue #166 again with some fresh eyes.  I propose the following changes:
Changes to Options:

* `--group-special`
  This is the same as `--special-card-groups` in the past.
* `--group-kingdom`
  This will group non-Kingdom (i.e., no randomizer) cards into an Extra section for the Expansion.
  This includes moving base cards to the Extra section if `--base-cards-with-expansion` is given.
*`--group-global [Events|Landmarks|Projects|Ways]`
  Combines the --exclude-xxx options into a single option that can contain any number of wanted card types to be included in the "Extra" section.
  This can be called once with any/all desired types, or called multiple times with one or more types.
  `--group-global` with no parameters will use ALL the global groups`.

  *QUESTION*: Should this stay as ALL, or should this be NONE (as if `--group-global` is not given.)

All three of these are independant and can be mixed.

The following are aliased and could be depricated if desired:
`--special-card-groups` aliased to `--group-special`
`--exclude-events`      aliased to `--group-global Events`
`--exclude-landmarks`   aliased to `--group-global Landmarks`
`--exclude-projects`    aliased to `--group-global Projects`
`--exclude-ways`        aliased to `--group-global Ways`

Changes to `types_db.json`:
Since all the global groups are special card types, added new fields to specify information that was hard coded in `main.py` before.
- Added "group_cost" to give the cost to display on the tab.  "*" for cards with various costs. "" for cards with no cost at all.
- Added "group_global_type" to indicate (a) the type for the resulting global group (b) indirectly, that this type can be grouped globaly.
Note: this has essentially eliminated the need for card.isEvent(), card.isLandmark(), etc.  Worked to make things generic via json entries.
Example:
```
    {
        "card_type": [
            "Event"
        ],
        "card_type_image": "event.png",
        "group_cost": "*",               <--- new
        "group_global_type": "Events",   <--- new
        "defaultCardCount": 1,
        "tabCostHeightOffset": -1,
        "tabTextHeightOffset": 0
    },
```

Changes to `cards.py`:
- for CardType(), added "group_cost" and "group_global_type" initialization (per above) and the corresponding `getGroupCost()` and `getGroupGlobalType()` methods.
- for Card(), added `get_GroupCost()` and `get_GroupGlobalType()` methods.

Changes to `sets_db.json`:
- Removed all "extras" entries.  These are now generated automatically in `main.py`.
- Added "has_extras" to designate if the set should have an Extra section.  Default is True, so only have to provide when False.
- Added "upgrades" which is used to indicate that this set upgrades another set.  Before, this was hard coded in `main.py`.  Now it is generic and is set in `sets_db.json`.
Example:
```
    "dominion2ndEditionUpgrade": {
        "edition": [
            "1"
        ],
        "image": "dominion2ndEdition_set.png",
        "has_extras": false,                    <--- new
        "upgrades": "dominion1stEdition",       <--- new
        "set_name": "*dominion2ndEditionUpgrade*",
        "set_text": "",
        "text_icon": "*"
    },
```
    
Changes to `/en_us/sets_en_us.json`:
Just needed to add a few entries for expansion extra sets that did exist previously.

Handling of most of these changes are in `main.py` and one line change in `draw.py`.

Also in `draw.py` changed `item.card.cardset.title()` to `" ".join(word.capitalize() for word in item.card.cardset.split())` to fix capitalization of 1St & 2Nd to 1st & 2nd (a pet peeve.)

### DIFF
--- a/card_db_src/en_us/sets_en_us.json
+++ b/card_db_src/en_us/sets_en_us.json
@@ -14,6 +14,11 @@
         "set_text": "There are strange things going on in your basement laboratories. They keep calling up for more barrels of quicksilver, or bits of your hair. Well it's all in the name of progress. They're looking for a way to turn lead into gold, or at least into something better than lead. That lead had just been too good of a bargain to pass up; you didn't think, where will I put all this lead, what am I going to do with this lead anyway. Well that will all be sorted out. They're also looking for a universal solvent. If they manage that one, you will take whatever they use to hold it in and build a castle out of it. A castle that can't be dissolved! Now that's progress.\nThis is the 3rd addition to <i>Dominion</i>.",
         "text_icon": "A"
     },
+    "alchemy extras": {
+        "set_name": "Alchemy Extras",
+        "set_text": "",
+        "text_icon": "A"
+    },
     "animals": {
         "set_name": "Animals",
         "set_text": "<i>Dominion: Animals</i> is a fan created expansion for the card game <i>Dominion</i>. It contains three new kingdom cards: \"Rabbits\", \"Yard dog\", and \"Gray Mustang\".\nSee https://boardgamegeek.com/boardgameexpansion/203184/animals-expansion-mini-fan-expansion-dominion.",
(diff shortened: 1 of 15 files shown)


wvoigt:
I still cannot run the `pre-commit`.  I'm having issues satisfying all the version constraints on my system.   I'm also not the best at git, so that could be part of it too.

sumpfork:
I think this sounds good in general, will try to find review code.

I also don't know how some of this code ended up on master due to a non-existent Travis build 1212. I've reverted that on master, but something's afoot.

sumpfork:
I just haven't had time to think about this, so as you two seem to agree this cleans things up I'm just going to trust you :).

Issue #123, opened 2016-11-29
Add functionality to save sets of options for later re-use


sumpfork:
I'd actually like to write a general library for argparse to do this for any app - I've run into this need several times. Barring that, just saving them to a JSON or config file would do.

sumpfork:
Actually, it looks like there's might be a more acceptable solution in https://github.com/bw2/ConfigArgParse since the last time I looked.

wvoigt:
I stumbled upon this independently recently. Looks like there have been recent updates on https://github.com/bw2/ConfigArgParse. See also https://pypi.org/project/ConfigArgParse/.

I did a little playing around. Can specify a config input file. Command line arguments overwrite the config input file values (good). Can specify a file to write resulting config (good.)
So far I have found two minor annoyances with the resulting write:

*If you specify a config file for input via a double --, that argument is included in the output, even though all the values from the file are also included. So if you used that output file as input, you would reload the original file, even though it is no longer needed.  If you just use a single -, then it is not included.
*The write exits immediately after the file is written. There is currently no option to run as normal after the write. Although, it looks like the right hooks are there to add this in the future.
    
The changes turned out to be minimal.  Because of the behavior above, just used -c for the config file and -w to write a config file.

wvoigt:
ConfigArgParse does seem to be new-ish.  Not clear how well it is maintained.

sumpfork:
The tech for this exists in the script since #314 and there's the above issues to expose them in the web generator, so closing this here in favour of those.

PR #314, opened 2020-04-06
Add ability to use and save options in files

wvoigt:
Created a pull request to make it easier to comment/review.


wvoigt:
This goes with Issue #123 

The failure seems to be the failure to load 'configargparse'.  I presume it needs to be installed.  Not sure if I can do that.

sumpfork:
You can - it needs to be added to `requirements.in` which then needs to be compiled to `requirements.txt` using pip-tools' `pip-compile`.

sumpfork:
Ah, sorry, I forgot. Run `doit compile_requirements` instead of `pip-compile`.

wvoigt:
OK, I'm at a loss.  The error is:
The command "python setup.py test" exited with 1.

That said, I could not run `doit compile_requirements` on my system for some reason.   Not sure if that changes anything in the background.

sumpfork:
Yeah, there's something still wrong with the travis setup. Master builds are failing because travis at some point checked in one of your PR branch changes into master - with a build number that doesn't exist in the travis build history. Not sure what's going on, need to find time to investigate. Has nothing to do with this PR as far as I can tell.

PR #363, opened 2022-01-26
Limit <line> to the text box width

### LINKED ISSUE #360: <line> overflows narrow card widths
When printing cards narrower than about 5 cm, the <line> graphic overflows onto a second line. Looks like the problem is the fixed width of this substitution:

        # <line>
        replace = "<center>{}</center>\n".format("&ndash;" * 22)
        text = re.sub(r"\<line\>", replace, text)

Instead of a fixed 22 en dashes, it should probably have a maximum width of 4 dashes times the text width in centimeters.

Alternately, the code for labels suggests that 4 cm is the minimum useful width for card text, so you could also change the output to a static 16 en dashes, which appears to be narrow enough for that width.

bszonye:
The `<line>` substitution expands to a string of 22 en dash characters, which is 11 ems wide. Any text box narrower than that will cause the `<line>` to break and overflow onto a second line. The `BodyText` font is 10 points, so the overflow happens with any text box narrower than 110 points (about 3.88 cm). Accounting for margins, that means that `<line>` breaks on cards narrower than about 5 cm wide.

This PR adds an `emWidth` parameter to the `add_inline_text` function to indicate the available horizontal space in ems. The function then limits the replacement to the number of em dashes that will fit. (The function previously measured in en dashes, which are exactly half the width of an em dash. Changing to em dashes eliminates an extra conversion step.)

Fixes #360.


sumpfork (approved):
Seems to work as advertised, thanks!

PR #373, opened 2022-03-25
Add Allies (no card text yet)

marksim:
This adds some of the basics of Allies including the wackiness with the split piles.

All of this was done by hand.


marksim:
Clearly do not know what I'm doing, but I'm hoping that some of this work is helpful for getting the Allies dividers up sooner rather than later.  :(

sumpfork:
Replaced by #376

@marksim: thanks for the contribution! All changes should be made in `card_db_src` and will propagate to the package's card db directory via the `doit` run. I tried to move your changes to the right places in #376 - I haven't actually looked at the outputs, but at least the tests pass now.

marksim:
Thanks so much.  Appreciate the info about how it works.

Issue #361, opened 2022-01-26
crop marks missing from partial pages


bszonye:
On the last page of a job, crop marks do not appear after the rightmost column unless there is:

* at least one full row on the page, or
* a horizontal gap between cards larger than the crop mark length

Likewise, crop marks do not appear below the bottom row unless every row is used, or there's a vertical gap between cards. This appears to be the consequence of this code in `draw.py`:

                    pageItems[i].cropOnTop = (y == rows - 1) or RoomForCropV
                    pageItems[i].cropOnBottom = (y == 0) or RoomForCropV
                    pageItems[i].cropOnLeft = (x == 0) or RoomForCropH
                    pageItems[i].cropOnRight = (x == columns - 1) or RoomForCropH

That only crops on the outside rows, which might be empty on the last page of a job.

wvoigt:
I am unable to test this out at the moment, but I think the following may do what you want:
```
                last_item = len(pageItems) - 1
                last_row = (rows - 1) - (last_item  // columns)
                pageItems[i].cropOnTop = (y == rows - 1) or RoomForCropV
                pageItems[i].cropOnBottom = (y == last_row) or (y == last_row + 1  and x > last_item % columns ) or RoomForCropV # bottom row, or next to bottom to the right of the last item
                pageItems[i].cropOnLeft = (x == 0) or RoomForCropH 
                pageItems[i].cropOnRight = (x == columns - 1) or (i == last_item ) or RoomForCropH # end of column or the last item
```

bszonye:
Thanks! That code looks reasonable to me. It's similar to how I would have solved the problem. I'll help out with testing, and if I get to it first I'll send a PR.

PR #390, opened 2022-05-21
Fix partial page crop-marks

wvoigt:
Add crop marks on partial pages. Fixes Issue #361


sumpfork:
@bszonye do you have time to test this? I don't at the moment...

bszonye:
The fix works for me, although there's one change I'd recommend if it's not too difficult to implement.

In the image, I've highlighted all of the new cropmarks added by the fix. The four marks in blue are good, but the two marks at the inside corner (highlighted in orange) overlap the diagonally adjacent card. That's not ideal, as it can leave a faint mark after cropping. If possible, it'd be better to suppress any marks originating from that inside corner.

Otherwise, LGTM. Thanks for the fix!

cropmarks

wvoigt:
> In the image, I've highlighted all of the new cropmarks added by the fix. The four marks in blue are good, but the two marks at the inside corner (highlighted in orange) overlap the diagonally adjacent card. That's not ideal, as it can leave a faint mark after cropping. If possible, it'd be better to suppress any marks originating from that inside corner.

I see what you are saying.  However, at this time, the cropmarks are controlled by the side of the card.  So Top, Bottom, Left, Right.  It will take a bigger code change to exclude just the Right upper and the Bottom left.   I need to think through the ramifications.

bszonye:
Cool, I wasn't sure how much hassle it would be to account for that subtlety. This is an improvement regardless. Thanks!

Issue #383, opened 2022-04-08
upgrade-with-expansion option clashes with the expansions option


md5i:
If you include `--upgrade-with-expansion`, all expansions with upgraded cards are output even if those expansions aren't included in a given `--expansions` option.  So, for example, if you do:
```
dominion_dividers --upgrade-with-expansion --expansions "intrigue"
```
It will include all the Dominion 1st and 2nd edition cards along with the Intrigue ones.

md5i:
I looked at creating a PR for this, but `--upgrade-with-expansion` is handled very early in `filter_sort_cards`, and `--expansions` is handled very late.  This part of the code is complicated enough that I don't know what I might break by moving things around.

wvoigt:
I lost my development/test system so I cannot verify this.  But I think the issue is here:
```
    # Combine upgrade cards with their expansion
    if options.upgrade_with_expansion:
        for card in cards:
            if Card.sets[card.cardset_tag]["upgrades"]:
                card.cardset_tag = Card.sets[card.cardset_tag]["upgrades"]
                options.expansions.append(card.cardset_tag.lower())         <----- problem here
```
If it is a card from an upgrade, then it first assigns the card to the expansion set it is upgrading.  Which is what you would expect for --upgrade-with-expansion.
But the last line then adds the expansion on to the list of expansions to print.  This should not be needed and is creating the wrong behavior.  This line should be removed.

md5i:
I traced it to this line as well but am hesitant to say more, as I doubt this was added for no reason whatsoever.  If I were to guess, it's so that if, say, Intrigue was selected, that with the `--upgrade-with-expansion` option would automatically include the "Intrigue 2nd Edition" expansion.  The implementation here is obviously flawed, but I don't know that *just* removing that line is correct.

wvoigt:
Why would you want "Intrigue 2nd Edition" to print when you upgraded the 1st edition?  This was clearly done under the false assumption that if this option was selected the person would want all the updated sets printed as well. As we are going to have more of these, I think that logic is false. We should let this option do as it says, and no more. Let the determination of what sets to print be handled by the other options.

md5i:
After thinking about it, I think you are right.

wvoigt:
[...]

PR #391, opened 2022-05-21
Fix issue with --upgrade-with-expansion adding expansions to print

wvoigt:
Fixed Issue #383.  Prior to the fix, in spite of any --expansion option, any expansion that had upgrades was printed. This fix correctly adds the cards to the correct expansion, and then removes the upgrade expansion from printing.


src/domdiv/main.py
```diff
@@ -1497,10 +1497,12 @@ def filter_sort_cards(cards, options):
 
     # Combine upgrade cards with their expansion
     if options.upgrade_with_expansion:
+        if not options.exclude_expansions:
+            options.exclude_expansions = []
```
sumpfork, line 1501: This is confusing, as it would assigned an empty list if it's already an empty list. Check should probably be `is None` instead of `not`.

sumpfork (changes requested):

src/domdiv/main.py
wvoigt: You are correct.  I will provide a correction.

Issue #387, opened 2022-05-06
Need some new split-banners


md5i:
For Allies, we need new split banner images for Treasure-Duration (Contract) and Duration-Victory (Stronghold).

Issue #386, opened 2022-05-06
grouping and card count incorrect for Allies


mr-ice:
When using --wrapper and --expansions=Allies I get most of the  Kingdom card wrappers out ok.

However, some of the new kingdom cards and special cards are different size sets.  For example the Augurs which have 16 cards.

When printing the wrappers, I expect these to be 1.6 times as thick as the regular wrapper.  They are closer to 3 (could be 3.2).

There are seven card sets where the wrappers are significantly oversized... 
Allies
Augurs
Clashes
Forts
Odysseys
Townsfolk
Wizards

The outstanding one is Allies, which is not one of the new 16 card sets but should be pretty normal special (not kingdom card) set.

Full options I use:
```
bin/dominion_dividers \
--minmargin=1.25x1.25 \
--no-page-footer \
--notch-length=1.30 \
--order=expansion \
--orientation=vertical \
--papersize=letter \
--size=sleeved \
--sleeved \
--special-card-groups \
--tab-side=right \
--tabwidth=4.75 \
--thickness=2.57 \
--wrapper \
--expansions="$@"
```

wvoigt:
You have --special-card-groups set. In the Clashes grouping there are 32 cards. You can confirm this by printing with  --count
I suspect it is similar for others.

mr-ice:
update: the Allies wrapper is the correct size (24 cards fit well)

16 augurs is larger than the Allies wrapper

Here are the Allies, Augurs, and Carpenter side-by-side

<img width="964" alt="Screen Shot 2022-05-05 at 9 34 08 PM" src="https://user-images.githubusercontent.com/1105176/167057438-3d434404-50fb-4687-9e7f-d43c8f183f2d.png">

mr-ice:
> You have --special-card-groups set. In the Clashes grouping there are 32 cards. You can confirm this by printing with --count
> I suspect it is similar for others.

Correct, without it the Allies don't generate a wrapper

Printing with --count, I do not see Clashes and other 4x4 sets showing 32 cards:

<img width="963" alt="Screen Shot 2022-05-05 at 9 44 58 PM" src="https://user-images.githubusercontent.com/1105176/167058698-d7bf5572-df43-4b85-98de-6e6fadd83222.png">

mr-ice:
however, turning off --special-card-groups does work for Clashes and Augurs, but --count in both cases shows 16:



[...]

PR #392, opened 2022-05-23
Fix Allies special cards, add banners, fix horizontal card consistency

wvoigt:
This does a several related things:
1. Fixes Issue #386.  Allies card counts should now be correct for Augurs, Clashes, Forts, Odysseys, Townsfolk, and Wizards.
2. Fixes Issue #387 by adding the two banners.
3. Makes the Ally cards consistent with other horizontal/landscape cards like Projects, Events, Landmarks, etc. Can now use `--group-global` to separate out if wanted.
4. Fixed up Boons, Hexes, and States to be consistent with other horizontal/landscape cards as well.  I did not change Artifacts, since those seem to always attach to a specific card.


md5i:
I think you typed in the wrong issue number for your first bullet point.

wvoigt:
Yep. Fixed.

sumpfork (approved):

Issue #359, opened 2022-01-25
request: option to select 2nd edition + removed cards


bszonye:
I have a full set of 1st and 2nd edition cards for the base game and Intrigue, and I keep all of the redundant cards in deep storage. Therefore, my main collection contains all of the 2nd edition cards and the 12 cards that were removed in the upgrade. I would like a way to select just those 12 cards without bringing in all of dominion1stEdition and intrigue1stEdition. Perhaps introduce two new sets similar to the 2ndEditionUpgrade sets, but reversed?

dominion1stEditionRemoved = Adventurer, Chancellor, Feast, Spy, Thief, Woodcutter
intrigue1stEditionRemoved = Coppersmith, Great Hall, Saboteur, Scout, Secret Chamber, Tribute

bszonye:
I see that I can use `--cardlist` to print just those 12 cards on their own, which is helpful. It would still be nice if I could select them directly with `--expansion` though, to include them easily in larger lists without getting two copies of all the cards that carried over to 2nd edition. If there's another good workaround, that would be fine too.

bszonye:
Even with `--cardlist`, I'm having a hard time selecting *only* the 2nd edition cards along with the removed cards. If I set `edition = all` then I get two versions of the cards that appear in both editions, but if I set `edition = latest` the script filters out *all* of the 1st edition cards, including the removed cards that I want to keep.

wvoigt:
Isn't the 2nd edition plus the removed cards the same as the 1st edition plus the upgrade (new) cards?

`--expansion dominion1stEdition --upgrade-with-expansion`
instead of
`--expansion dominion2ndEdition`
My understanding is that for the dividers, the card texts are the same for both additions (the latest with errata).  The only difference are the cards contained in each set.  The only difference I can think of right now is that the set icon for the common cards will be 1st edition instead of 2nd addition.

bszonye:
Thanks for the feedback! Unfortunately, it doesn't help because I don't want the upgrades and removed cards sorted together. I keep the "removed" 1st edition cards separately, along with the promo cards. So far, the best method I've found for this is to print the 1st edition cards in a separate batch, with a `--cardlist` to select only the 12 cards. Every other approach that I've tried either produces two copies of the common cards or sorts them with the 1st edition cards (which affects tab positions).

[...]

PR #393, opened 2022-05-24
Added sets for removed cards from 1st editions of Dominion and Intrigue

wvoigt:
This fixes Issue #359.
Adds sets for the cards that were removed from the first edition of Dominion and Intrigue.


sumpfork:
Hmm, I haven't tried, but with this if no options are specified are the cards included both as standard and as removed? If so, I think we should start having some sets that are excluded when nothing is specified?

wvoigt:
As is, it adds another set with the specific removed cards.  They also appear in the 1st edition set.  This is very similar to how the upgrade sets are printed even though they are also included in the 2nd edition sets.

That said, I agree the default prints a lot of sets & cards now. It would not be too hard to add on to the empty `options.expansions` check and add "*Removed" to the `options.exclude_expansions`.  Might make sense to have this as a constant list so it is easier to add more to it in the future if needed. Something like: 
```
DEFAULT_EXCLUDE_SETS= ["*Removed"]

...

    if options.expansions is None:
        # No instance given, so default to all Official expansions
        options.expansions = ["*"]
        # But exclude the seldom used expansions		<-- New
        if options.exclude_expansions is None:			<-- New
            options.exclude_expansions = []			<-- New
        options.exclude_expansions.append(DEFAULT_EXCLUDE_SETS)	<-- New
```
If you think this is a good idea, I can easily add it.

sumpfork:
Yeah, I think by default as little as possible unexpected card duplication should occur, and second editions should be preferred.

wvoigt:
Another way to do it is just change `--edition` to be 'latest' instead of 'all' when no expansions are listed. It would be simpler and have the same result.

wvoigt:
This is what I will implement.  When no expansions are specified:

- If no `--edition` is specified, then the new default will be to print all the latest editions (no 1st editions, no upgrades, no removed cards).  Prior behavior printed everything.
- If `--edition` is specified, all editions will be filtered by the `--edition` parameter (so it over rides the above).

The prior default behavior can be done by either specifying `--expansions *`  or  `--edition=all`

To give easy options to just print upgrade cards or removed cards, I will add `upgrade` and `removed` as options to `--edition`.

bszonye:
Thanks, this sounds great! Looking forward to trying it out.

Issue #389, opened 2022-05-18
Prosperity Second Edition


nickv2002:
Hot on the heels of [Seaside Second Edition](https://github.com/sumpfork/dominiontabs/issues/364) we're also getting [Prosperity Second Edition](https://boardgamegeek.com/boardgameexpansion/362517/dominion-prosperity-second-edition) sometime in the June 2022 timeframe.

> Dominion: Prosperity (Second Edition) contains nine new types of Kingdom cards, which will be released on their own as an update pack for those who own the first edition of this expansion. A few cards will receive minor errata, but these will not appear in the update pack.

![Prosperity Second Edition Box Art](https://cf.geekdo-images.com/yyHAfwOS1hBhTUKbD18B-A__imagepage/img/g0OOChwip4wnNZZ5770MQS3H3d4=/fit-in/900x600/filters:no_upscale():strip_icc()/pic6849008.jpg)

nickv2002:
Prosperity 2nd Edition Previews are complete. [The new and removed cards are detailed on the Dominion Strategy Wiki](http://wiki.dominionstrategy.com/index.php/Prosperity#Kingdom_cards)

wvoigt:
Changes in Pull Request #396.
I could use a 2nd set of eyes to verify this looks correct.
[ProsperityTest.pdf](https://github.com/sumpfork/dominiontabs/files/8838370/ProsperityTest.pdf)

PR #396, opened 2022-06-04
Changes for Prosperity 2nd Edition

wvoigt:
Separated out 1st and 2nd Editions, added Upgrade cards as well as Removed cards.
This should fix Issue #389.


wvoigt:
I could use a 2nd set of eyes to verify this looks correct.
[ProsperityTest.pdf](https://github.com/sumpfork/dominiontabs/files/8838370/ProsperityTest.pdf)

sumpfork:
Landed as #404 after merge and db update.

Issue #408, opened 2022-06-26
Hinterlands 2nd Edition


nickv2002:
Looks like the Second Edition updates will continue with Hinterlands 2nd Edition releasing as soon as July 2022. Should follow the model established so far with 9 new cards and a separate upgrade pack. [Dominion Wiki Details](http://wiki.dominionstrategy.com/index.php/Hinterlands)

![Hinterlands 2nd edition box art](https://cf.geekdo-images.com/S8nAwjjJYIzqveU-qdc8WA__medium/img/sLUjvX3KgEMe1aMj5HLLFcmpDgk=/fit-in/500x500/filters:no_upscale():strip_icc()/pic6924579.jpg)

wvoigt:
Pull request #409 created with the changes.
For those who would like to proof read, here is the a printout with the 1st edition, 1st edition removed, 2nd edition, and 2nd edition upgrade cards.
[Hinterlands.pdf](https://github.com/sumpfork/dominiontabs/files/9033409/Hinterlands.pdf)

databoy01:
Just checking in with the status of this. Thank you everyone for the hard work too.

nickv2002:
@databoy01 I think it's ready to merge other than needing a verification pass on all the PDFs that @wvoigt posted above (something I have not had time to do). 

If you (or anyone else) do read over the PDFs that were posted, just leave a comment that they look good or if you found any issues. Thanks!

databoy01:
Hello Nick, I receive my upgrade pack tomorrow, once I receive it I will verify the cards that come with it to the PDF. If I find any issues I’ll let you know. Thanks

nickv2002:
Sure thing. Thanks for any help. For others still waiting for their cards you can see [scans of them all on the Dominion Wiki](http://wiki.dominionstrategy.com/index.php/Hinterlands)

sumpfork:
The PR (#409) is actually merged so you can already use the script on master for these. I just haven't updated the online generator.

databoy01:
Hello Peter @sumpfork - Not sure how to do that, so I guess I will just wait until its uploaded.  Thank you for making this amazing script and program.

sumpfork:
Ok, this is deployed.

PR #409, opened 2022-07-02
Hinterlands 2nd Edition changes

wvoigt:
Fixes Issue #408.  This adds the Hinterlands 2nd edition, removed cards from the 1st edition, and the upgrade pack.


wvoigt:
For those who would like to proof read, here is the a printout with the 1st edition, 1st edition removed, 2nd edition, and 2nd edition upgrade cards.
[Hinterlands.pdf](https://github.com/sumpfork/dominiontabs/files/9033409/Hinterlands.pdf)

PR #417, opened 2022-08-18
Feature German 2nd Edition Prosperity

Xedorian:
I included the German translations for the new prosperity cards and their rules as well as clarified two promo card wordings.


sumpfork:
Hi, it looks to me like you're not resolving changes from the master branch correctly. Your current PR reverts recently landed changes from master, both in the DB files and in the test setup (that's why the tests are failing). You need to accept changes from master when merging them in.

Xedorian:
Ok, thanks. I'll try to find where I can allow that^^

PR #421, opened 2022-09-27
Update hinterlands2ndEdition Set Icon

nickv2002:
Make icon match official one better
[As suggested on BGG](https://boardgamegeek.com/thread/926575/article/40642111#40642111)


nickv2002:
Should be an easy/direct update. Please merge if looks good!

PR #426, opened 2022-11-22
Poetry setup

sumpfork:
Toying with the idea of using this...


sumpfork:
I updated the build system recently but decided against poetry.

PR #427, opened 2022-12-27
GZip in-package card db

sumpfork:
Having both the source and target of the card database be plain JSON leads to no end of confusion from contributors. This gzips the one that should not be manually modified.

In the process, it also lazy-loads the files so that the database can be compiled without already existing in some form.


nickv2002 (approved):
👍 to this change generally. I didn't do a detailed review but looks good at a high level.

sumpfork:
There'll be some work in making the online generator work with this due to the constant removals, but I think it's worth it.

Issue #420, opened 2022-08-31
Dominion: Plunder Expansion


nickv2002:
Well no end in site for new Dominion cards. RGG just announced **Dominion: Plunder** ([BGG](https://boardgamegeek.com/boardgameexpansion/369536/dominion-plunder))

> Across the sea, they have so much stuff. And it's so much better than your stuff. Finer craftsmanship. Better quality materials. Shinier. They have crowns, tiaras, and diadems — and that's just the hats. It's time to get some of that stuff. You want an easy life, and you're prepared to work hard for it, so you've rounded up some old salty dogs, plus a sourpuss and a bitter goldfish. And set sail. The sea is a harsh mistress, but a good cook, at least if you like everything really salty. There are red skies tonight, so they'll be making a batch of Sailor's Delight, which you understand to have tuna fish in it. And soon you'll be attacking merchant ships and taking their treasure. But the real treasure is the happy memories you'll be making.
> 
> Dominion: Plunder, the fifteenth expansion for Dominion has lots of Treasures and Durations, with cards that give you Loot, and Traits that modify piles. Events return.

Release TBD, presumably by the end of 2022.
![pic7035290](https://user-images.githubusercontent.com/704392/187744407-5a2696c4-e1d9-4cec-a619-bc98b5a5ffa6.jpg)

nickv2002:
The [official rulebook](https://www.riograndegames.com/wp-content/uploads/2022/08/DomPlunder.pdf) and [Dominion Strategy wiki Page](http://wiki.dominionstrategy.com/index.php/Plunder_(expansion)) (with text of all cards) are up now.

databoy01:
Is it done yet…lol. Look forward to this expansion!

md5i:
Well, I got my physical copy yesterday.

md5i:
So, I am working on this.  I'm writing some scripts to scrape the raw data from the Dominion Strategy Wiki.  I need a reminder, though.  Am I supposed to be editing the files in `card_db_src` or the files in `src/domdiv/card_db`?

md5i:
Also, is there a noticeable difference between `<n>` and `<br>` in output?  Under what circumstances should one be used versus the other?

sumpfork:
> So, I am working on this. I'm writing some scripts to scrape the raw data from the Dominion Strategy Wiki. I need a reminder, though. Am I supposed to be editing the files in `card_db_src` or the files in `src/domdiv/card_db`?

[...]

PR #430, opened 2022-12-30
Add Plunder set.

md5i:
This has a placeholder traits.png file based on way.png.  It should be replaced with a proper image for Traits.


md5i:
Associated with Issue #420.

sumpfork:
Thanks! Will try to get to review this weekend or Monday.

sumpfork:
Looks really good to me and seems to work as far as I can tell. 

Thanks again!

PR #433, opened 2023-01-03
Replacement for #413

sumpfork:
This is a merged version of #413 after the Plunder update. No other updates, but I've been testing @bszonye's changes and they seem to work. I'm also adding a minimal set of of radio buttons to the generator for the `meta` versions of the new params.


src/domdiv/draw.py
```diff
@@ -1486,7 +1486,7 @@ def drawTab(self, item, panel=None, backside=False):
             # adjust dimensions based on the application image metrics
             bannerHeight += cardType.getTabTextHeightOffset() * tabScale
             # adjust for space around banners and scalloped edges
-            margin = tabWidth / 18 if card.isExpansion() else tabWidth / 48
+            margin = tabWidth / 18
```
bszonye, line 1489: I don't think this change works as intended. The margin is the distance from the edge of the tab to the artwork. The `tabWidth / 18` setting pushes the artwork inward so that it doesn't clash with the scalloped edges on the expansion dividers. The normal banner tabs have more room, so the `tabWidth / 48` pushes the artwork closer to the edge. Changing it to `tabWidth / 18` actually leaves _less_ room for text.

If you want to leave more room for text, the place to change that is the `padding` variable, which is initialized a few lines above:
```
# whitespace
margin = padding = 2
```

bszonye (commented):
Most of the changes look like they're to unrelated code, except for one change to the tab margins which I've just tested & found a problem. I don't think that change is doing what it's intended for, based on the commit log. I left more details in a code comment.

src/domdiv/draw.py
bszonye: I'm curious which case prompted this change. The minimum 2mm padding _should_ prevent the text from ever running into the artwork. The only exceptions that I know are about are cases where the text is too wide to fit, even after breaking it into two lines and shrinking to the minimum allowed font size. In those cases, this change will make things worse because the margin goes outside the artwork, and `tabWidth / 18` is wider than `tabWidth / 48`.

sumpfork: Maybe I misinterpreted, but before this change the text was running into the left side of the artwork on the banners if I turned off the cost icon, and always was for dividers that don't have a cost (Traits, for example). It was a quick fix with me trying to interpret old and changed code seeing how `margin` was used, so feel free to suggest a better one. Looked ok to me after the fix for the settings combinations I tried.

sumpfork:
I guess I should be clear: it's not that the text is running into the left side of the tab, but some of the artwork as features like bolts (again, check Traits) that look bad when overlapping with the text.

sumpfork:
In particular, without change:
Screenshot 2023-01-02 at 10 48 51 PM

with change:
Screenshot 2023-01-02 at 10 50 42 PM

bszonye:
Oh, I see! I tested extensively on the common banner graphics, but I missed that problem with the landscape frames because I usually print centered text. The previous version handled the expansion dividers well but obviously not the landscape cards.

I think it's better if the frames with room can use the wider margins, but that might require an additional attribute in `types_db.json` to handle it in a non-hacky way. Now that this is merged, I'm preparing a followup PR with some graphical improvements, so I might handle it there.

Thanks for the quick response!

sumpfork:
Sounds good, and yes, I think an additional offset may be needed - I looked for that first, but it seems we don't have a horizontal one right now.

bszonye:
Awesome! Really happy to see this released. Happy new year! 🎉

src/domdiv/draw.py
bszonye: Oh! Your comment here inspired me to look at all of the graphics with & without the cost icons, and I actually like your margins better for everything _except_ the cost icons. That's good because it means that it should be easy to adjust without any further type attributes. Instead, the cost icons can outdent a bit when they're combined with artwork. I'll fix that in my next PR. Thanks!

bszonye: On further investigation, the 1/18 margin is quite close to the card art! If anything, the card margins are a bit wider on the coin side. The banners are 52.5mm wide, and domdiv adds a 1mm cushion around that for a total of 54.5mm. If we made a tab the same width as the cards, 1/18 = ca. 3mm from the edge of the tab, or 2mm from the edge of the banner.

The coins on the cards are set 2.5mm from the left end of the banner, and the set icons are set 1.5mm from the right end, and honestly the set icons are a bit cramped on the cards. Therefore it actually makes sense to use a margin 1/18 of the tab width! I don't know where I got the 1/48 value from earlier. I wonder if I was looking at an outlier, or perhaps the graphics from one of the digital adaptations?

There's one small flaw with the `tabWidth / 18` calculation. It scales the margin with the entire `tabWidth` even though there's always a fixed 1mm cushion around the artwork. It should be something like this instead:
```
    margin = 1 + (tabWidth - 2) * 2.5 / 52.5
```
I'll fix that in my followup PR.

bszonye:
Took a quick look at the generator interface, that looks good to me too. Nice!

Issue #414, opened 2022-08-15
all treasures use base card graphics


bszonye:
I've noticed that all of the Treasure cards use the same yellow banner graphic, which comes from the Dominion base cards (Copper, Silver, Gold, Platinum, Potion). Most treasures use a different graphic. Ideally there could be different types for kingdom Treasure cards and base supply Treasure cards.

I believe the Victory cards have a similar problem, except that all of them use the kingdom graphics instead of the base supply graphics, such that Estate, Duchy, Province, and Colony are the ones with mismatched graphics.

bszonye:
I have a partial fix for this here:
https://github.com/bszonye/dominiontabs/compare/bradd/wrappers...bszonye:dominiontabs:bradd/banners

It's only a partial fix because it changes *all* of the treasure banners to use the kingdom-style graphics. Ideally, the base cards and kingdom cards should each use their own style of banner. With this WIP, only the Curse cards use the base supply style, while the base Treasure and Victory cards use the kingdom style. There's also a little hack to put white text on the Night cards, which should ideally be handled in the types database and not via heuristics.

Also, this fix is dependent on my work in #413 so it's not really ready until that's further along anyway.

bszonye:
With #436 merged, all treasure and victory cards use the standard kingdom card banner. It would be nice to use the special base card banners for just the nine base cards (Copper, Silver, Gold, Platinum, Estate, Duchy, Province, Colony, Curse). However, given how the card type system assumes a 1:1 relationship between card types and graphics, that change might be moderately difficult. I'm content with the current state of things, so please feel free to close this issue.

Issue #415, opened 2022-08-15
Ways graphic has different proportions from other landscape cards


bszonye:
Most of the landscape cards have graphics with narrow rectangles like this:
![image](https://user-images.githubusercontent.com/5016349/184635872-b6074382-ae9f-4c2a-b91b-0a1c709cc028.png)

There's a transparent area at the top of the graphic that keeps it from filling up the whole tab, and the text is centered appropriately for that. The `way.png` graphic is different from the rest. It has a taller rectangle without the transparent spacer at the top:
![image](https://user-images.githubusercontent.com/5016349/184636565-0432578a-5cfd-4b43-b89b-8e0485cd6e7d.png)

The end result is that the Ways print differently, with a taller rectangle that doesn't match the other landscape cards, and the text ends up off-center in the rectangle.

PR #436, opened 2023-01-03
Improve tab artwork & layout

bszonye:
Several improvement to the tab banners:

* improved horizontal layout per comments on #433
* re-scanned card graphics at higher resolution
* calibrated colors with an X-Rite color checker
* adjusted coin/debt/potion vertical alignment to match the new scans
* changed Night banners to use white text on black like the cards do
* closed the big gap where the banners wrap around a flagpole in the card art
* changed Treasure cards to use the kingdom banner instead of the base card banner

Most of the changes update the graphics to resemble the card art more closely. The last change addresses #414:

Dominion uses two different graphics for Treasure and Victory cards. The eight base types (Copper, Estate, etc) use a wider banner than non-base cards like Gardens or Quarry. The current version uses the special base card graphics for _all_ treasures (including e.g. Quarry) but not for any of the victory cards. This set of graphics standardizes on using the same shape for all treasure and victory cards.

Ideally, the base treasures and victory cards would have a separate type using the alternate graphics, along with a similar rescan & touch-up for both types + Curse, but that's very labor-intensive work that I haven't had bandwidth for.


sumpfork:
Looks great, my only nitpick is that now the set icons on event-style banners run into the right artwork on them:
Screenshot 2023-01-03 at 2 50 31 PM

bszonye:
I noticed that too. (I paid close attention to the landscape banners this time!) I have a few thoughts about it and a couple of possible solutions.

First, here are all of the landscape banners at the default 4.0cm tab size:
image

For comparison, here they are at 5.3cm, which makes action banners "actual size" (same scale as the physical cards):
image

The frames on the original cards have different aspect ratios. Because the `draw.py` renderer stretches artwork to fit the tab, this results in all of the rivets looking stretched or squeezed. Events, landmarks, and projects all end up with wide oval rivets, while all of the post-Nocturne landscapes get narrow ovals. (Ways have an additional problem documented in #415.) The event banners clash the most because they have the largest, darkest rivets.

My first reaction is that the landscape artwork needs a rescan and standardization similar to this PR, using a single image recolored appropriately to each landscape type. Given the limitations of the current renderer, I would adjust the artwork so that the rivets look perfectly round at 5.3cm "actual" size. That would leave plenty of room for icons at both ends. I don't have the resources to do a good job of this, unfortunately, as I only have access to the Adventures and Nocturne landscape cards.

My next observation is that the Dominion cards themselves have this problem with the set icons. It's especially noticeable with the Adventures cards, and Coin of the Realm is the worst:

image

The cards aren't totally consistent, but most of them have about a 2.5mm left margin and a 1.5mm right margin (relative to the widest points of the banner). That puts the set icons into the shadowy part of the banner, which is fine on some card types and unreadable on others. I don't like that either, plus it's extra bad on the landscape cards, so I decided that it was best to use the same margin on both sides.

Still, **I agree that the margins are a bit too tight for events, landmarks, and projects.** My preferred solution would be to improve the graphics rather than to special-case the margins, which is why I didn't address it further in this PR. However, if you want it to look better _now,_ I have a couple of fixes in mind.

Last night, I noticed that all of the landscape cards (and only the landscape cards) have a `group_global_type` attribute in the cards db, which is exposed via the `Card.get_GroupGlobalType` method. We could use that information to implement a new interface to test for all landscape cards as a group:
```
    def isLandscape(self):
        return self.getType().getGroupGlobalType() is not None
```
which then lets us adjust the landscape margins en masse:
```
              # fit text inside banner/frame graphics
              margin = (tabWidth - 2 * safety) * marginRatio
              if card.isLandscape():
                  margin *= 1.5
```
image

This has some negative consequences, however. I don't like the way it brings the landscape symbols _slightly_ out of alignment with the regular cards, and I think it looks especially bad on the post-Nocturne landscapes. (I tried a few different values for the margin adjustment, and x1.5 was the least bad, in my opinion.)

I think it looks a lot better with a x1.75 adjustment on _just_ the three types that need it:
```
              # fit text inside banner/frame graphics
              margin = (tabWidth - 2 * safety) * marginRatio
              if card.get_GroupGlobalType() in ("Events", "Landmarks", "Projects"):
                  margin *= 1.75
```
image

I don't love the special-casing here. This sort of thing would be better handled in the card db, perhaps by giving each type a `margin_ratio` attribute appropriate to its artwork. However, that's a much bigger change, and if we're going to make a big change, I think it would be _much_ better to just standardize the artwork to all work with the same margins.

In summary, I propose one of these resolutions:

1. fix the artwork so that landscape cards all look consistent and work with `marginRatio = 2.5 / 52.5`
2. adjust only the events, landmarks, and projects to have x1.75 margins
3. adjust all landscape cards to have x1.5 margins

My first choice would be to do **2 now** and **1 later.** My second choice would be to do nothing now and 1 later. I think 3 actually makes things worse, although it might be handy to have the `Card.isLandscape` method regardless. Opinions?

bszonye:
One other thing: I'm not sure what to do with the source PSD file for the new graphics, which is _285 MB.__

I had a similar problem with the `coin.psd` in #413. I assembled the coin graphics from a collage of all the treasure cards with the numbers cut out. The original has a ton of layers and weighs 931 MB, so for check-in I merged most of the layers to get it under 6 MB. I still have the original, I just didn't want to through a whole gigabyte at your repo.

The `banner.psd` file isn't amenable to that kind of flattening, so I'm not sure what to do with the source file. Please let me know what you'd like me to do about it. Also let me know if you'd like the original `coin-source.psd` file & if so how best to transfer the data to you.

bszonye:
I pushed a commit for my preferred resolution, in case you agree with my conclusions.

nickv2002:
It probably makes sense for us to save those PSDs with the project in some way. [Git/GitHub Large File Storage](https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-git-large-file-storage) is probably the right tool for the job if you can manage the CLI interface.

bszonye:
GitHub LFS does make sense, but the monthly free bandwidth (1 GB) is quite low relative to the file sizes. That's not even enough to hand off both original Photoshop files (285 + 931 MB) in the same month. I'd like to hand off a copy of the original image sources, but I don't really have the (metaphorical) bandwidth to learn the LFS tools & deal with the potential costs right now.

sumpfork:
I'll try to provide a place to upload these files for you on the weekend and review this PR - given the thought you've clearly given it I'm likely to just land what you did though :).

bszonye:
I forgot to mention it in the PR description, but this also fixes #439. The original banners for duration-victory and treasure-duration have blank coins embedded in them. I noticed that when preparing the new versions, but it didn't occur to me to report it as a bug.

PR #303, opened 2020-03-21
feat: updated sets and translations for bigbox 2 german edition

despian:
A couple of cards in the German Bigbox v2 have slightly different translations.

I added 2 new sets (`promo-bigbox2-de` and `guilds-bigbox2-de`) to support the different names of the cards:

"Carcassonne" -> "Befestigtes Dorf"
"Wahrsager" -> "Hellseherin"

PR #440, opened 2023-01-07
German BigBox

sumpfork:
Updated replacement for #303.


sumpfork:
Landing this as it'll be hard to accept after the sorting of the source JSON files.

PR #441, opened 2023-01-07
Dutch translation for Allies

Antheii:
Dutch translation for Allies, hope I did correctly with those new gzip files

PR #448, opened 2023-01-09
Replacement for #441

sumpfork:
- run pre-commit
- compile card DB


sumpfork:
Replaces #441 though I'm finding that hard to review as many cards moved around in the db files.

PR #452, opened 2023-01-09
Add --tab-artwork-resolution option to limit DPI.

### LINKED ISSUE #443: Generated file size is too big
After @bszonye's changes to improve the banner images the filesize for a complete divider file jumped from <5mb to 24mb. That's ok for local printing, but because of the way the online generator works I'm limited to a 6mb response size (imposed by AWS lambda), so even generating a single expansion throws an error with the new version.

A couple of options I can think of:
- scale the images down to a size closer to the original one (sad to lose the print quality)
- have two versions and an option to switch to smaller sizes for the online generator, but still distribute the bigger ones for local script usage
- Instead of putting the pdf in the response and having the browser download it, put it on S3 and provide a download link instead
- The above S3 solution, but combine with the two size solution and generate both, return the smaller, also point to the bigger
- The preview also is failing. It only generates the first 2 pages, so I'll have to look into why even that size is too big. I think it's because we actually spec the whole file and then later limit it to two pages?

I don't have time to rework the backend to not be Lambda-based and want to keep it this way for cost reasons anyway.

bszonye:
Fixes #443.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -7,4 +7,4 @@ testpaths = tests
 test = pytest
 [flake8]
 max-line-length = 120
-ignore = E203,W503
+ignore = E203,W503,B006
--- a/src/domdiv/draw.py
+++ b/src/domdiv/draw.py
@@ -1369,9 +1369,51 @@ class DividerDrawer(object):
                 w += pdfmetrics.stringWidth(part[1:], font, small)
         return w
 
-    def drawTab(self, item, panel=None, backside=False):
+    artwork_cache = {}  # cached images from the drawArtwork method
+
+    def drawArtwork(self, image, x, y, w, h):
         from io import BytesIO
 
+        opacity = self.options.tab_artwork_opacity
+        dpi = self.options.tab_artwork_resolution
+
+        # Get the image from cache if possible.
+        cache = (image, x, y, w, h, opacity, dpi)
+        render = self.artwork_cache.get(cache)
+
+        if render is None:
+            # Get the original image.
+            render = DividerDrawer.get_image_filepath(image)
+            # Make any optional adjustments.
+            if opacity != 1.0 or dpi:
+                imgObj = Image.open(render)
+                if opacity != 1.0:
+                    # Set image opacity.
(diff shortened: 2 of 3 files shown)


setup.cfg
```diff
@@ -7,4 +7,4 @@ testpaths = tests
 test = pytest
 [flake8]
 max-line-length = 120
-ignore = E203,W503
+ignore = E203,W503,B006
```
bszonye, line 10: FYI, B006 is the flake8-bugbear code for using mutable data structures as default parameters:
```
def bugbear_dislikes_this(sequence=[]):
    ...
```
My dev environment has flake8-bugbear enabled by default & I was tired of it complaining about this. It wouldn't be a bad idea to add bugbear to the project tools & clean up the thing it's complaining about, but I figured that was out of scope for this change.

sumpfork (approved):
That's fine, but why not a "get_adjusted_artwork" or similar helper that uses a functools cache decorator instead of the explicit cache implementation?

bszonye:
I moved away from @functools.cache because it isn't safe to use on instance methods, and three of the important inputs are instance attributes (opacity, resolution, and canvas). However, I just spotted a mistake in my cache key, so don't merge yet.

bszonye:
Oops:
```
AttributeError: module 'functools' has no attribute 'cache'
```
That decorator was added in Python 3.9, and I suppose that means the pipeline is using an older Python version. No big deal, I can switch to @functools.lru_cache instead, but that's something to keep in mind if you're planning to use any newer Python features.

bszonye:
My earlier "oops" was because I included the x & y coordinates in the cache key, even though those parameters don't contribute to the artwork adjustments. I figured as long as I needed to fix the cache anyway, I might as well also try to get rid of the manual cache management. Gimme a sec and I'll fix the new failure.

bszonye:
I'm done messing with it now (unless something else goes wrong). Please let me know if you have any further feedback. The suggestion to refactor was a good one; I wasn't happy with the manual caching code.

PR #456, opened 2023-02-22
German translation of missing kingdom cards

JustusDinera:
I translated all Plunder and Allies kingdom cards. I also translated missing ones from Hinterlands 2nd edition.


sumpfork:
Needs #459.

sumpfork (approved):
Thanks, seems to work once I generated the actual package.

PR #464, opened 2023-05-24
Translated some card names to Spanish

InigoGutierrez:
This commit adds translations to Spanish for card names for the following sets:

*  Dominion 2E
* Intrigue 2E
* Seaside 1E
* Dark Ages

I just wanted to create the tabs, so I didn't translate the text or rules for the cards, just the names. I also translated card types and I tried to translate the bonuses but the `doit update_languages` command would not generate the `.gz` for it so it is translated in the `.json` but I just copied and renamed the `en_us` bonuses `.gz` file for the program to work.

If this looks like a random collection of sets, well, they are the ones I have and wanted tabs for. Great tool! I really appreciated all the options of the CLI.


sumpfork:
Thanks! I'll have to look into the problem you're describing, hopefully on the weekend.

One note: you should be able to leave out any of the English entries like "description" and the script should fill them in with the English version. This might be preferable as it's easier to see what's not translated yet, and the text stays up-to-date with the English version until the entries are actually translated.

InigoGutierrez:
Oh, thanks for the input, I might update the pull request then, if I find it not too cumbersome to empty all the text I left in English.

sumpfork:
Accepting as is for now after testing, can clean up later.

PR #468, opened 2023-05-29
Cleaned Spanish translation files

InigoGutierrez:
I finally cleaned the files so they fall back on the english version for the cards and text I did not translate.


sumpfork:
You inspired me to do a more general automatic cleanup: #469.

PR #469, opened 2023-05-29
Cleanup language dupes

sumpfork:
This
- introduces a script to be manually run that strips untranslated entries from card translation files. It
  - removes "description" and "extra" if either of them is the same as in English
  - remove "name" only if it's the same as in English and neither "extra" nor "description" are translated
  - remove the whole card entry if nothing is left after the above
 - allows cards to not exist in English but only in other languages
   - this is specifically to allow the German big box extra cards to not have to have an entry in English
   - to do so, empty cards from the card db do not get empty defaults anymore (I think this was bad anyway as the card tag is not guaranteed to be a printable card name)
   - it is not allowed for a card to not be present in the English version, and the usual override from other languages will still apply
   - the actual generation does not fall back on English anymore, but rather only loads the translated text as these should have been augmented with English as necessary during the card db generation step

Issues:
- this strips entries that are the same in two languages, for example a `description` of `+1 Coin` (as this gets replaced with a coin symbol. I think that's ok, but it does make it a little less obvious as to what's translated at first glance.
- you can still generate an "expansion" like the German box in English, and it will silently not have the German-only cards. Not a big deal I think, but it would probably be better to fail with a good error here.


nickv2002:
Nice work. Sounds like a good simplification overall.

nickv2002 (approved):

PR #502, opened 2024-02-21
Add other platforms to run tests on

sumpfork:
Add other platforms to test on in github actions as we've had trouble with differing behaviours.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py38
+envlist = py39
 
 [testenv]
 deps = pytest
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py38
+envlist = py39
```
sumpfork, line 2: unrelated (and unused in CI) change, just matching python version required elsewhere.

nickv2002 (approved):
LGTM. Pretty neat that GitLab supports all these OS versions.

PR #505, opened 2024-02-26
update dockerfile to get fonts from GDrive instead

nickv2002:
This improves the `dockerfile` that was trying to be too clever about fonts. Instead it just downloads them from a zip file on Google Drive.

---
The features below were moved to another PR:

> This also adds an initial workflow to build docker image and upload to [GitHub Container registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry)
> 
> Some GPT generated suggestions for setting it up (I don't think I have access to create the tokens)
> > Log in to GitHub Container Registry: Using the docker/login-action, we log into GHCR using the credentials stored in GHCR_USERNAME and GHCR_TOKEN. This step is critical for allowing subsequent docker push commands to authenticate against the GitHub Container Registry.
> > 
> > Build and push Docker image: We build the Docker image using the docker build command. The tag (-t) for the Docker image is set to include the GitHub Container Registry URL (ghcr.io), the repository name (${{ github.repository }}), a specific name for your application (change myapplication as necessary), and the commit SHA (${{ github.sha }}) for uniqueness. Afterwards, the docker push command uploads the built image to GHCR.
> > 
> > Make sure to:
> > 
> > Replace myapplication with the desired name for your Docker image.
> > Adjust any Docker build arguments or context as necessary for your specific Dockerfile.
> > This integration adds a robust CI/CD pipeline step for containerization alongside your existing Python package deployment, streamlining both code and container distribution with each release.


nickv2002:
@sumpfork this is waiting on your review since I made a change to GitHub workflows. If I should take that out and just update the dockerfile LMK.

dockerfile
```diff
@@ -1,28 +1,39 @@
-FROM --platform=linux/arm64 pacodrokad/fonts:latest AS fonts-image
+FROM alpine AS fonts-download
+# get the fonts from Google Drive
+RUN wget "https://drive.usercontent.google.com/download?id=1_BmJ1afnSt1rR_YAWhUCFDch8Bat78ti&export=download&authuser=0" -O fonts.zip
```
sumpfork, line 3: This is a personal google drive? I really don't like depending on that.

sumpfork (changes requested):

sumpfork:
> @sumpfork this is waiting on your review since I made a change to GitHub workflows. If I should take that out and just update the dockerfile LMK.

Yeah, let's take that out and try it in a separate PR.

dockerfile
```diff
@@ -1,28 +1,39 @@
...
+# Copy the requirements file first, to cache the requirements layer as well.
 COPY requirements.in .
-RUN pip-compile --no-emit-index-url requirements.in && \
-    pip install -r requirements.txt
 
-# Copy the local directory contents into the container at /app
+# Now compile and install dependencies in a single layer
+RUN pip-compile --no-emit-index-url requirements.in \
+    && pip install -r requirements.txt \
+    && rm -rf ~/.cache/pip
```
sumpfork, line 31: I don't think this should run in the dockerfile as it fails to pin the requirements like it's supposed to. Instead, if we need a requirements file specific to the docker image (why do we?) we should generate a requirements file under a different name using this command in the docker image and copy it to be `requirements.txt` for this docker image.

dockerfile
```diff
@@ -1,28 +1,39 @@
...
+
+# Copy the rest of your application code. This step is done later as this part changes more often.
 COPY . .
 
-# install the application
+# installation using setup.py
 RUN python setup.py develop
```
sumpfork, line 37: I don't think this needs to be `develop`? Why not `pip install .`?

dockerfile
nickv2002: I understand where you're coming from. The normal solution would be to add them to this repository but that's not possible. 

I'm trying to make it easier to run the CLI tools: "use the docker image" usually works well for many other open-source projects I use daily, so I want to build the fonts into the image. Is there some other place we could link these to enable Docker to grab them when it's building?

dockerfile
nickv2002: I was just following instructions from https://github.com/sumpfork/dominiontabs?tab=readme-ov-file#developing
`pip install .` will probably work too, I'll check.

sumpfork: Yeah, except this is for running, not developing. Unless I guess this is meant to be a dev container too, but then I think devs can figure out to change this line if they want an editable version installed.

dockerfile
nickv2002: I'm not sure I follow what you're saying here.

Can we just use the `requirements.txt` from the main repo, then just `pip install -r requirements.txt` (followed by removing cache)?

I didn't do that before simply because of the note in the `requirements.txt` that it was auto generated.

nickv2002: changed to `pip install -r requirements.txt` without `pip-compile` beforehand

dockerfile
nickv2002: Changed to `pip install .` as suggested rather than following dev instructions

dockerfile
nickv2002: I've resolved the other items but would appreciate your thoughts here on how we can make the dockerfile easy to use.

sumpfork: I just don't think we can distribute them legally, so we shouldn't try.

I was cleaning requiements stuff up #507 anyway so cleaned up the dockerfile in a similar manner to this PR over there. Also added an example as to how to map the local dir into the container so you can use local fonts and actually get the pdf back out easily.

nickv2002:
This was handled in https://github.com/sumpfork/dominiontabs/pull/507

PR #507, opened 2024-03-23
Cleanup requirements, docker file

sumpfork:
- get rid of `requirements.in` in favour of directly compiling `pyproject.toml` into `requirements.txt`
- add `dev-requirements.txt` in the same way, based on the dev requirements in `pyproject.toml`
- require a reasonable python version > 3.7
- simplify the docker file to
  - not pip-compile anything (it shouldn't, those versions are pinned for a reason)
  - not try to handle fonts (we can't distribute them)
  - not run `--help` by default - behaviour should be the same as the normal script
  - update the README to show how to map a the local dir so you can actually get the output file and use local fonts


dockerfile
```diff
@@ -1,29 +1,15 @@
...
-COPY requirements.in .
-RUN pip-compile --no-emit-index-url requirements.in && \
-    pip install -r requirements.txt
-
 # Copy the local directory contents into the container at /app
 COPY . .
 
 # install the application
-RUN python setup.py develop
+RUN pip install .
```
nickv2002, line 13:
```suggestion
RUN pip install . && rm -rf ~/.cache/pip
```

nickv2002 (approved):

PR #510, opened 2024-04-10
Improve logging; simplify font handling

sumpfork:
Use loguru for logging. Turn down log level by default but add option to set it.

Clean up font use in the non-latin1 case.

@smrky1

### DIFF
--- a/src/domdiv/cards.py
+++ b/src/domdiv/cards.py
@@ -1,6 +1,7 @@
 import json
 import re
 
+from loguru import logger
 from reportlab.lib.units import cm
 
 
@@ -208,11 +209,7 @@ class Card(object):
                     setImage = Card.sets[self.cardset_tag]["image"]
 
         if setImage is None and self.cardset_tag != "base":
-            print(
-                'warning, no set image for set "{}", card "{}"'.format(
-                    self.cardset, self.name
-                )
-            )
+            logger.warning('no set image for set "{self.cardset}", card "{self.name}"')
         return setImage
 
     def setTextIcon(self):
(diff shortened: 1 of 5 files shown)


src/domdiv/cards.py
```diff
@@ -208,11 +209,7 @@ def setImage(self, use_set_icon=False):
                     setImage = Card.sets[self.cardset_tag]["image"]
 
         if setImage is None and self.cardset_tag != "base":
-            print(
-                'warning, no set image for set "{}", card "{}"'.format(
-                    self.cardset, self.name
-                )
-            )
+            logger.warning('no set image for set "{self.cardset}", card "{self.name}"')
```
nickv2002, line 212: Nice upgrade, no need for f-strings I assume.

src/domdiv/draw.py
```diff
@@ -586,100 +586,90 @@ def registerFonts(self):
...
-            pdfmetrics.registerFont(
-                TTFont(
-                    "Times-Italic-TTF",
-                    pkg_resources.resource_filename(
-                        "domdiv", fontpaths.get("Times-Italic-TTF")[0]
-                    ),
-                )
-            )
+        timesTTFs = set(["Times-Roman-TTF", "Times-Bold-TTF", "Times-Italic-TTF"])
+        timesTTF_not_found = timesTTFs - set(fontpaths)
```
nickv2002, line 598: 👍 good use of `set`

nickv2002 (approved):
Nice upgrades to logging and readability of the code.

src/domdiv/cards.py
sumpfork: Actually no, I think this is wrong :). From the docs I initially thought there was no need, but there is. The lib can do `format`-style substitution but then needs the args supplied explicitly, so I went with standard f-strings instead. I'll fix this.

Issue #498, opened 2024-02-06
Rising Sun, Marchland, 2nd Edition Cornucopia & Guilds


nickv2002:
FYI and future TODOs that new cards are incoming in the next few months.

- A new promo [Marchland](https://wiki.dominionstrategy.com/index.php/Marchland)
- A new expansion [Rising Sun](https://wiki.dominionstrategy.com/index.php/Rising_Sun)
	- Donald X says "My best guess is March 2024, but as usual I will know more in say February." 
- A 2nd edition for the [combined Cornucopia & Guilds expansion](https://wiki.dominionstrategy.com/index.php/Cornucopia_%26_Guilds)

nickv2002:
Looks like we're getting a 2nd edition of Cornucopia & Guilds in march and the cards have already been revealed:
https://wiki.dominionstrategy.com/index.php/Cornucopia_%26_Guilds

I'll update the title of this issue to reflect that.

thale13:
Any updates on this?  Just wondering.  Willing to help if you can point me in the right direction.

nickv2002:
@thale13 I imagine we'll fold the promo and Corn/Guild 2nd Ed. updates into a Rising Sun update once those cards are known (that's how it's worked in the past). Not sure when that will be but probably soon. [Watch the Dominion Wiki for more info on that](https://wiki.dominionstrategy.com/index.php/Rising_Sun).

If you do want to get started you can add cards to the JSON card DB, eg Marshland would get a new entry [somewhere around here](https://github.com/sumpfork/dominiontabs/blob/master/card_db_src/cards_db.json#L7290-L7299).

dbgalur:
This second edition / upgrade combines the Guilds and Cornucopia sets into a single set.  So we can talk about cards that were removed from the 1st edition of each, but the new cards are added to a combined set, rather than to a second edition of each individual set.  How would you represent the upgrade set in sets_db.json?  Specifically, "upgrades" expects a string rather than an array, no?

dbgalur:
Actually, that may not be a relevant question.  5 of the new cards are clearly part of Cornucopia, while the other 3 new cards are part of Guilds, even if they are not explicitly marked with different icons.

nickv2002:
We obviously try to match what has been physically printed because they're dividers for physical cards. So we'll have an upgrade set and a second edition set and a first edition set (see previous upgrade packs here). The upgrades and second edition will be for the dual pack, but the first edition remains separate.

[...]

PR #516, opened 2024-05-04
Add Cornucopia and Guilds second edition and upgrade pack.

dbgalur:


sumpfork:
Thanks! I'll try to find time to have a stab at trying it out this weekend. Probably just needs some minor formatting cleanup (`pip install pre-commit` would do cause this to be done when you `git commit`) so the linters are happy.

dbgalur:
I'm trying to debug an issue where the Tournament and Rewards are not showing up as a group if I try to output the "cornucopiaAndGuilds2ndEdition" expansion with "--special-card-groups". (If I output "cornucopia2ndEditionUpgrade" or do not try to group special cards, it does work.).

sumpfork:
#498

sumpfork:
Replaced by #517

PR #517, opened 2024-05-04
Cornucopia and Guilds 2nd Edition

sumpfork:
Replacement of #516


dbgalur:
I didn't add a new set icon!

nickv2002:
> I didn't add a new set icon!

We usually use [icons from here](https://boardgamegeek.com/filepage/73273/dominion-card-icons-vector-images) but that doesn't have a Corn/Guilds 2nd Ed update yet either. I'm watching the thread on BGG so I can add when they do have an update.

PR #518, opened 2024-05-13
Add missing translation for all available german cards and sets

Jano1312:
Updated about 150 cards and most set descriptions in the German translations.
All available cards in German should now be translated.


sumpfork:
Replaced by #522

PR #524, opened 2024-06-12
Add Marchland promo.

md5i:


md5i:
Looks like I didn't do something that I needed to do when adding the card?  What did I miss?  I ran update-languages.

(Better documentation for adding card entries would be appreciated.)

sumpfork:
The failing test tries to tell you (perhaps in an obscure way):
```
assert len(cards) == num_cards_expected
E       assert 1018 == 1017
E        +  where 1018 = len(["Alms", "Amulet", "Artificer", "Ball", "Bonfire", "Borrow", ...])
```
So because you added a card the test needs to be updated with the expected number of cards (1018).

I'm not sure how useful this test is, but it does make us check that the card count is what we intended, or didn't change if we didn't intend to change it.

md5i:
Is that something that I, when submitting a pull request, should update when adding a card?  I'm happy to add that to this PR, if so.

sumpfork (approved):

PR #527, opened 2024-08-17
Add Cornucopia & Guilds set icon

### LINKED ISSUE #520: Cornucopia/Guilds 2e set icon
The new second-edition cards for Cornucopia and guilds still use the old set icons - would it be possible to get this updated so that they use the new one? Thanks!

sumpfork:
Closes #520


card_db_src/sets_db.json
```diff
@@ -81,7 +81,7 @@
             "upgrade"
         ],
         "has_extras": false,
-        "image": "cornucopia_set.png",
+        "image": "cornucopia_guilds_set.png",
```
nickv2002, line 84: @sumpfork I think the Upgrade Set should get the upgraded icon too. 
(I didn't rebuild the `sets_db.json.gz` because I'm a limited connection but I think the build scripts do that)
Otherwise looks good. Thanks.

nickv2002 (approved):

card_db_src/sets_db.json
nickv2002: Actually, I got `doit update_languages` to work and uploaded that `sets_db.json.gz` so I think it's all good now.

PR #535, opened 2024-09-24
Rising Sun setup

sumpfork:
Starting a Rising Sun PR based on https://github.com/sumpfork/dominiontabs/issues/498#issuecomment-2369800260

Things I added:
- preliminary images
- test adjustments


spazard1 (approved):
LGTM.

src/domdiv/images/sun.png
nickv2002, line 1: I added this image which is just [referred to as a Sun or a Sun Token](https://wiki.dominionstrategy.com/index.php/Prophecy#Official_Rules).

src/domdiv/draw.py
```diff
@@ -1193,6 +1193,8 @@ def replace_image_tag(
             # Potion
             (r"(\d+)\s*\<\*POTION\*\>", "potion_small.png", 2, 140, 1.5),
             (r"Potion", "potion_small.png", 1.2, 100),
+            # Sun
+            (r"Sun", "sun.png", 1.2, 105),
```
nickv2002, line 1197: I added the rule to replace it in card text (though I'm not sure if that would impact anywhere else yet). There are a few other cards that have "sun" in their text...

sumpfork: I wonder whether it shouldn't be `SunToken` or `RisingSun` instead of `Sun` to avoid accidental use?

nickv2002: I checked a few cards with "sun" in their names and this seems to work OK, so approving. Keep an eye on it though.

nickv2002 (approved):

src/domdiv/draw.py
nickv2002: I'm fine with changing it to `SunToken` Let me do that....

nickv2002: Done, changed to `SunToken`

TistoH:
Thank you all for the work involved in integrating this into the main db and creating new images.  I only wish I could test it (my research says that a tool dependency prevents me from using cygwin as my development environment).

nickv2002:
> Thank you all for the work involved in integrating this into the main db and creating new images. I only wish I could test it (my research says that a tool dependency prevents me from using cygwin as my development environment).

If you can use docker images you might try those. There's [some docs on the homepage](https://github.com/sumpfork/dominiontabs#docker) that work fine on this `rising_sun` branch to compile and output PDFs.

PR #537, opened 2024-09-29
Add nl_NL translations for Plunder expansion

siblingsofthevoid:
Help, I think i just lost my sanity...

In all seriousness, these are all card descriptions and extras as per Dutch cards and manual.
In basically all cases i tried to preserve the order as it is in the en_US translations.

PR #541, opened 2024-10-01
Siblingsofthevoid add plunder nl nl translations

sumpfork:
Replaces #537


sumpfork:
These should all be deployed to the online generator now.

Issue #544, opened 2024-10-17
pip version of 4.8.0 missing pkg_resources dependency


dekarguy:
Wanted to try out a handful of options that aren't on the online generator and followed the instructions and am consistently being met with the following error

```
C:\Users\dekar>dominion_dividers --help
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\Users\dekar\AppData\Local\Programs\Python\Python313\Scripts\dominion_dividers.exe\__main__.py", line 4, in <module>
    from domdiv.main import main
  File "C:\Users\dekar\AppData\Local\Programs\Python\Python313\Lib\site-packages\domdiv\__init__.py", line 1, in <module>
    from pkg_resources import DistributionNotFound, get_distribution
ModuleNotFoundError: No module named 'pkg_resources'
```

I'm on windows 11, python 3.13.0, ran pip install domdiv and then tried running with --help to get the options

I haven't installed any of the fonts, expecting it would rely on fallbacks, and the instructions for the font installation seem to be missing a step as well, the readme mentions running a package install below, but there are no package install instructions in the readme itself

sumpfork:
Yeah, I think I had this report somewhere before on windows. We should move off pkg_resources, there's an issue for that #325. Can you try just installing it manually?

The "package install below" is under development in the readme, pip install -e .[dev]

dekarguy:
I wasn't able to install pkg_resources, but pip install setuptools ended up working, so a note about that in the instructions for python 3.12+ would be useful :)

PR #545, opened 2024-10-17
Remove use of pkg_resources

### LINKED ISSUE #325: Update resource packaging and access
We should use the builtin https://docs.python.org/3.8/library/importlib.html?highlight=importlib#module-importlib.resources instead of the external setup used right now.

sumpfork:
This has come up a couple of times, latest in #544.

pkg_resource from setuptools is deprecated and often not installed (we don't specify it as a runtime dependency anywhere, which is bad.

So, upgrade to use importlib.resources.

Fixes #325.

### DIFF
--- a/src/domdiv/db.py
+++ b/src/domdiv/db.py
@@ -1,6 +1,5 @@
 import copy
 import functools
-import importlib.resources
 import json
 import os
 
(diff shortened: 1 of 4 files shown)


nickv2002 (approved):
Seems reasonable, this is the kind of thing that if the checks pass we can ship with little concern. :shipit:

PR #563, opened 2025-05-13
Keep arrows in card names for groups, but always render in Times Bold

mbbush:

This adds a workaround for the fact that the arrow character isn't present in the various custom fonts used on dominion cards, by always rendering it using Times Bold. I'm removing a different workaround that had instead replaced it with a dash.


sumpfork:
Small is much preferred, greatly increases the chance I'll get to them in reasonable time.

sumpfork (approved):

PR #564, opened 2025-05-13
Options for multiple set icons

mbbush:

This PR includes new options for:
* Including multiple set icons on the expansion dividers. When playing with a mixture of 1st edition and 2nd edition cards, I find having both the 1st edition and 2nd edition icons on the expansion divider helps find which section to put the cards away in. I even put all 8 promo icons on the Promo divider, which fits fine even in vertical orientation 
image
  - This could probably use a bit more testing to see how it interacts with all the sets with multiple editions, and perhaps some better logic for deciding when to include both icons (e.g. if you only have 2nd edition cards, skip the 1e icon). But I didn't need that for my purposes so haven't implemented it yet.
* Grouping removed cards together with the 2nd edition. This is complementary to the existing option for grouping upgrade cards with the 1st edition. The differences are which set icon appears on the cards which are common to both editions, and which edition's text is used for the expansion divider.


src/domdiv/main.py
```diff
@@ -281,6 +281,11 @@ def filter_sort_cards(cards: list[Card], options) -> list[Card]:
             if Card.sets[card.cardset_tag]["upgrades"]:
                 options.exclude_expansions.append(card.cardset_tag.lower())
                 card.cardset_tag = Card.sets[card.cardset_tag]["upgrades"]
+    if options.removed_with_expansion:
+        for card in cards:
+            if Card.sets[card.cardset_tag].get("removed", False):
+                options.exclude_expansions.append(card.cardset_tag.lower())
```
mbbush, line 287: This will conflict with the change from list to set in #565. Either this PR or #565 can be merged first, but then the other one needs a minor update before it can merge.

mbbush:
```suggestion
                options.exclude_expansions.add(card.cardset_tag.lower())
```
If #565 is merged first, this is the necessary change.

PR #565, opened 2025-05-13
Refactor option parsing to use sets

mbbush:

This PR is just some cleanup. The most significant change is removing the second call to `clean_opts`, which looks like it was added only for the unit tests. I moved the call into the tests instead. 

The problem with `clean_opts` is that it assumes that options like `expansions` are a list of lists, and flattens them, but since in python a string is iterable, if it's called a second time, it turns 
```['allies', 'plunder']```
into 
```
['a', 'l', 'l', 'i', 'e', 's', 'p', 'l', 'u', 'n', 'd', 'e', 'r']
```

Fortunately most (all?) of the code which is called after `calculate_layout` doesn't care about the expansions in the options, but there's no guarantee of that.

### CODE REVIEW

**Comment by mbbush** [comment]
Yes, I think making `clean_opts` idempotent is a better solution. I've done that now. It ended up being a decently large refactor, changing the type of the various collection options from list to set. The tests were quite helpful, including new ones I added.

File: tests/generation_tests.py [hunk 1]
```diff
@@ -32,3 +34,45 @@ def test_grouped():
...
+        "trait",
+        "way",
+        "prophecy",
+        "boons",
+        "projects",
+        "event",
+        "boon",
+        "landmark",
+        "ally",
+        "allies",
```
mbbush, line 62: Why are both of these here? The test passes, so this is what the code is doing, but this doesn't seem right to me.

mbbush: hmm, actually, they're all duplicated, it's just not clear because this list is in whatever order it was in when I copied it from my debugger, not alphabetized. I'll clean that up, at least.

src/domdiv/config_options.py
```diff
@@ -834,6 +834,25 @@ def parse_opts(cmdline_args=None):
...
+            # The arg parser sometimes generates a list of lists
+            out.update(flatten_lower_and_deduplicate(item))
+        elif isinstance(item, set):
+            out.update(item)
+        elif isinstance(item, str):
+            out.add(item.lower())
+        else:
+            logger.warning(
+                f"Ignoring configuration option {item} with type {type(item)} as it has an unexpected type."
+            )
```
sumpfork, line 852: I feel like this should either add the item (so work for things that aren't strings) or throw an exception here?

src/domdiv/config_options.py
```diff
@@ -908,79 +927,55 @@ def clean_opts(options):
...
-        options.group_global = []
-    elif not any(options.group_global):
-        # option given with nothing indicates all possible global groupings
-        options.group_global = db.get_global_groups()[1]
-    else:
-        # options.group_global is a list of lists.  Reduce to single lowercase list
-        options.group_global = [
-            item.lower() for sublist in options.group_global for item in sublist
-        ]
+    legacy_groups = set()
```
sumpfork, line 957: Not in this PR, but we should deprecate these. There are more than enough options already.

sumpfork (changes requested):

src/domdiv/config_options.py
mbbush: Amen. I just don't know what's being used by the website, so it's hard for me to know what can and can't be removed.

src/domdiv/config_options.py
mbbush: Updating it to throw an exception.

src/domdiv/config_options.py
sumpfork: You can see it here: https://github.com/sumpfork/bgtools_web/blob/main/assets/lambda/domdiv_form.py#L30 - it's pretty custom, and I'll likely have to update that anyway when I deploy all these changes. I had been working to fully automate representing all relevant options of the script on the site directly (as an 'advanced' tab), but it ends up being somewhat cumbersome, so this is what it is for now.

sumpfork (approved):

PR #566, opened 2025-05-14
Allow sorting non-kingdom cards to the end of the expansion

mbbush:
I want to be able to put my events, landmarks, etc at the end of the cards in an expansion, with alternating tabs. This lets me do that. 

I don't fully understand the purpose of the "extras" pseudo-expansions. Maybe they're for this same purpose? If there's another way to accomplish this, I'd be happy to close this PR and instead update the documentation to make it more clear.

### DIFF
--- a/src/domdiv/config_options.py
+++ b/src/domdiv/config_options.py
@@ -30,7 +30,7 @@ SPINE_CHOICES = ["name", "types", "tab", "blank"]
 
 EDITION_CHOICES = ["1", "2", "latest", "upgrade", "removed", "all"]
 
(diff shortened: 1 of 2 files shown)


sumpfork:
Can you see whether `--group-kingdom` does what you want? It puts all kingdom cards first per expansion, then all other cards.

mbbush:
That's similar, but not exactly the same. For one thing, it prints extra expansion dividers (I'm using --expansion-dividers). It also resets tab ordering as opposed to continuing from where the expansion left off. So I still think this approach is valuable.

I really liked how easy it was to add a new way of ordering.

sumpfork (approved):

PR #572, opened 2025-05-18
Consistently capitalize card tags

mbbush:
I'm working on a big project to update all the card text using the dominion strategy wiki. It's mostly done, and looking great so far. Expect a PR soon, but it will be a big one.

I'm trying to pull out what small pieces of that work I can separate to get them in first. 

In this PR, I am:
* Updating the card_tag values to consistently match the name of the card in English, at least for cards that are present on the dominion strategy wiki. The only card tags remaining that are not capitalized are the card "groups" like "adventures events". Since that's internal to this code I see no reason to change it. It makes the work of automation far simpler when I can reliably assume that the card tag is equal to the name of the wiki page for that card. 
* Added two prophecies from Rising Sun. That's as far as I got before I switched from manual copy-pasting to automating everything. I plan to add all the rest in my big PR.
* Made a couple minor typo fixes and errata changes. Again, most of these will change again in the big PR, but there's no harm in including these few minor improvements.

You may find it easiest to review each commit one at a time. The re-alphabetize commit only contains changes made by the pre-commit hook (I committed the un-alphabetized files using `git commit --no-verify` so the diff would be clear). 

As far as when this merges relative to my other PRs, I can handle whatever order ends up happening, as far as managing the git conflicts.


nickv2002:
These changes seem generally reasonable but potentially could cause corner case issues their name slugs new characters (like spaces) are used elsewhere. Maybe we just merge this at the same time as the "update all the card text using the dominion strategy wiki" PR instead of changing things twice?

mbbush:
My preference would be to merge this first, as it will conflict with the changes in the big PR and otherwise make it harder to review. 

I don't entirely understand what you mean about corner cases. Right now allies is the only expansion that uses lower_snake_case for its card tags; all the others have Title Case Names With Spaces.

As far as the number of changes, @sumpfork said he generally preferred more, smaller changes than fewer large ones. Not every commit needs to be deployed to the website generator.

sumpfork:
> As far as the number of changes, @sumpfork said he generally preferred more, smaller changes than fewer large ones. Not every commit needs to be deployed to the website generator.

I can guarantee that not every commit will be deployed :)

sumpfork (approved):

nickv2002:
Oh gotcha, my oversight there, it wasn't apparent when just looking at the diff.

> I don't entirely understand what you mean about corner cases. Right now allies is the only expansion that uses lower_snake_case for its card tags; all the others have Title Case Names With Spaces.

Yeah, merging before deploying online as Sumpfork and you suggested is a good idea here.

PR #580, opened 2025-05-23
Fix set tags

### LINKED ISSUE #573: translations Guild/Cornucopia expansion tab
Hi, 

[Gilden.pdf](https://github.com/user-attachments/files/20271938/Gilden.pdf)
[Overvloed.pdf](https://github.com/user-attachments/files/20271937/Overvloed.pdf)

2 weeks ago i created dutch (NL) dividers of Guild and Cornucopia and noticed that the expansion tab did not translate.
can you check this please.?

thanks and keep up the good work

Gino

mbbush:
There are three changes here:
1. Changing "dark ages" to "darkAges" everywhere. This matches the convention used for all the other expansions, and avoids the need to learn the details of unix shell quoting to enter it on the command line. I did the replacement programatically and I'm confident I found everything.
2. Update the set tags for guilds and/or cornucopia in several non-english languages to refer to the first edition instead. 
3. Set a short name when missing for guilds and cornucopia in Dutch and German. It looks like because of the way we first load the english data, then overload the language data, the short name from english was being chosen in favor of the long name from the selected language. 

Fixes #573


sumpfork (approved):

sumpfork:
Mind rerunning `update_languages` on this?

PR #589, opened 2025-07-02
Fix a few French translations

Snaptraks:
I created tabs for my games (in French) and noticed a few discrepancies with text in the game documents and some special grouped cards (the Urchin - Mercenary -> Orphelin - Mercenaire and Hermit - Madman -> Ermite - Fou) had their text in English instead of French.

This should add the correct information for cards, taken from the individual card information in the cards_fr.json file.

The missing translations for the sets (sets_fr.json) are taken from my games' documents and rulebook.

I do not own all the expansions and different editions so those are the only ones I can verify, sorry!


sumpfork:
Thanks! If you can set up for it by installing the dev dependencies and running `doit update_languages`, then checking in the resulting .gz file updates?

Snaptraks:
> Thanks! If you can set up for it by installing the dev dependencies and running `doit update_languages`, then checking in the resulting .gz file updates?

Perfect, it was my understanding that those files were supposed to be generated upon PR/merge. I have included them now!

sumpfork (approved):

PR #608, opened 2026-01-19
Basic pyton native fontfixer | .otf -> .ttf

eddeT:
Basic python native fontfixer using cu2qu and fonttools. Converts .otf fonts to .ttf. 

Could not see any glaring issues with the fonts converted this way when scrolling through a full generation of all cards.

About the conversion:

- Converting outline format: Uses cu2qu to convert PostScript cubic bezier curves to TrueType quadratic curves
- Replacing font tables: Removes CFF/CFF2 tables and adds glyf/loca tables (TrueType format)
- Fixing font signature: Changes sfntVersion from 'OTTO' (OpenType/PostScript) to '\x00\x01\x00\x00' (TrueType)
- Updating maxp table: Changes version from 0.5 (CFF) to 1.0 (TrueType) with proper TrueType-specific fields

**To get conversion working I got help from Claude Sonnet 4.5.**

This will be something extra to maintain in the end, so perhaps not worth to add, just wanted to throw it out there.

How to test:

- Put .otf fonts in local_fonts
- Run ```uv run --extra fontfix fontfix ```
- Run ```uv run dominion_dividers --font-dir local_fonts```
- Provided correct fonts are converted in local_fonts, no font warnings should be raised
- Check the generated .pdf. It should contain expected fonts.


sumpfork:
I'm fine with merging this if it works. Let me know and mark as ready?

eddeT:
Tried it on two different setups and created pdfs from it. If anyone else could try as well that would be even better!

sumpfork (approved):

PR #611, opened 2026-02-06
Fixed typos

ecm85:
Courtyard: Missing S in "Cards"
Laboratory: Extra '.'
Ghost Ship: Missing S in "Cards"
Develop: Missing word "coin"
Swashbuckler: Missing '.'
Vassal: Missing '.'
Souk: Fix dash (it was the only one using it, I *think* it was an em-dash and the rest weren't, but they're all consistent now)
Despration: Fixed double space
Secret Chamber: had a lower-case c in '+X Cards'
Pirate ship: Fixed text (no version has ever had Coffers in it, that i can tell)
Tournament: Removed prizes at the end (they're also in "Tournament and Prizes" and didn't belong here, I think)

### DIFF
--- a/card_db_src/en_us/cards_en_us.json
+++ b/card_db_src/en_us/cards_en_us.json
@@ -796,7 +796,7 @@
         "name": "Courtier"
     },
     "Courtyard": {
-        "description": "+3 Card<br>Put a card from your hand on top of your deck.",
+        "description": "+3 Cards<br>Put a card from your hand on top of your deck.",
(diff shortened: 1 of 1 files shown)


ecm85:
Not sure what exactly the standard process is here. I don't have a python env set up atm. Is PR 612 sufficient to accomplish that task?

sumpfork:
> Not sure what exactly the standard process is here. I don't have a python env set up atm. Is PR 612 sufficient to accomplish that task?

Yup, for an otherwise clean PR like this it's easy for me to just land both.

sumpfork (approved):

PR #620, opened 2026-04-01
Updating empires split piles to work like allies split piles. Fixed one missing period in Groom description.

ecm85:


sumpfork (approved):
enough messing with github actions, sorry

